PR #14, opened 2021-02-22
First draft of model abstraction

phipsgabler:
A draft of what I have described in https://github.com/TuringLang/AbstractPPL.jl/discussions/10#discussioncomment-374373.


phipsgabler:
In the first step, I have written down the results of @yebai's and my brainstorming [here](https://github.com/TuringLang/AbstractPPL.jl/blob/phg/model_abstraction/interface.md) (and the distillation of previous discussions with @cscherrer).  The gist of it is:

- Models are mainly defined by their ability to `sample` and evaluate a `logdensity`.  This somewhat corresponds to `Distribution <: Sampleable` from Distributions.jl.
- These functions are generalizing over what DynamicPPL.jl and AbstractMCMC.jl are currently doing: `sample` should work the same way for posterior approximation using MCMC, but also include, e.g., exact forward sampling from generative models.  This should be possible through dispatch.
- Instead of having generalized semantic functions `logprior`, `logjoint`, etc., this interpretation should be encoded in the model type.
- To abstract both DynamicPPL.jl and Soss.jl,  we are thinking about two main types of models: generative and conditioned ones.  These should be separate types, between which you can switch with "transformations" `condition` and `decondition`.
- Modelling languages have wildly varying "trace types": named tuples, `VarInfo`, `ChoiceMap`, ... So we had this idea of using a common "probability expression" type here (the `@T` macro, name still in flux) to specify an assignment of random variables as required by the interface (conditioning and density evaluation).  This is inspired by [StatsModels.jl's `@formula` notation](https://juliastats.org/StatsModels.jl/stable/internals/#Internals-and-extending-the-formula-DSL): the expression can be written down and handled in a model- and PPL-independent way, with a "schema application" step for concrete instances.

What do you say?  CC @devmotion @cpfiffer @torfjelde 

(Not that at this stage we are still at "what is it that defines a PPL *in general*", and not yet at the step to factor out much concrete code from DPPL.  Although DPPL and Soss served as the main constraints to what the design should look like.  A free-standing `DensityModel` would be a good little example instantiation, though.)

phipsgabler:
One thing I forgot: "model" here means "probabilistic program" -- a more specific thing.  The idea is for this to be a subtype (or sub-interface) of `AbstractMCMC.AbstractModel`.

trappmartin:
Cool, I guess it's not required that a model has a log density, right?

devmotion:
Some initial thoughts:
- I don't think it's completely correct that in Turing models are conditioned with fixed observations: the raw model is always the same function, regardless of the values of the observations and whether they are present or not (e.g. if they are `missing`), and hence the `@model` macro does not condition on anything in a strict sense. Even in the `Model` instances we just happen to bundle some data in addition to this general model definition but you don't necessarily have to use the bundled data when executing the model. So even if model instances might look like conditioned models in Turing, actually we always deal with the generative model (otherwise, e.g., the `prob` macro wouldn't work).
- The Distributions.jl interface is based on `rand` and `logpdf` (and `loglikelihood`) but not `sample` and `logdensity`? My impression is that `rand` would be a better fit here, I think typically `sample` is used in the StatsBase sense of sampling from a collection with or without replacement etc. (and actually `AbstractMCMC.sample` is a bit of an abuse of this method). Also I don't think a `logpdf` is necessary for a model, similar to samplers in Distributions.jl it should be fine to just sample from them?
- The term `logdensity` is a bit unclear - is it the joint probability in the sense of LogDensityProblems? Is it fine if it is defined only up to a constant (if yes, is there an API to retrieve the actual log joint probability)?
- In general, I think it is important, whenever possible, to extend appropriate methods in the ecosystem instead of defining AbstractPPL-specific versions, in particular when one uses common names such as `sample` (StatsBase), `loglikelihood` (StatsBase), `logpdf` (Distributions), or `logdensity` (LogDensityProblems?)
- Wouldn't it make sense to encode the interpretation of `logprior`, `logjoint`, etc in a context that can be provided at execution time instead of having to define a new model instance every time? And maybe even provide shortcuts such as `loglikelihood(model) = logdensity(model, LogLikelihoodContext())`?
- The `@formula` syntax has never grown on me (neither in R nor in Julia) :smile: 
- I would prefer more descriptive names than `@T` and `@P` :slightly_smiling_face:

cscherrer:
There are some interesting ideas here, but I think I need to understand the high-level context better before I'll be able to address the details in any helpful way.

Is it correct to say that the problem this is addressing is that similar concepts appear across PPLs, so it's useful to encode that commonality? If this work is as successful as possible, how does the ecosystem come to look different?

I think this is especially tricky with respect to Soss. A Soss model has an "initial representation" - it stores the AST provided by the user, and provides some tools for mapping this to "primitives" like `rand` and `logdensity` that are useful for inference.

Given this, the biggest potential I can imagine for commonality is something that provides a back-end compilation target. Is that the idea here? What can be done to ensure we can connect this in a way that gives the advantages of all systems without adding unnecessary constraints, and without a situation where `n` PPLs become `n+1`?

Project.toml
```diff
@@ -7,7 +7,9 @@ version = "0.1.4"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
+StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
 AbstractMCMC = "2, 3"
+StatsBase = "0.33.4"
```
devmotion, line 14: What was the reason for adding the StatsBase dependency? Was it missing?

phipsgabler: Ah, hm. TBH I wouldn't have merged this yet... It is an accidential remainder from earlier things I tried.  

But still, this package would specify `StatsBase.sample` as part of the interface, so it probably should depend on and reexport that, right?

devmotion: `StatsBase.sample` is reexported by AbstractMCMC, so it's not necessary to depend on StatsBase directly to use and/or reexport `StatsBase.sample`. In fact, it might be easier to manage dependencies and versions if AbstractPPL does not depend on it if this would be the only reason.

phipsgabler: Cool, then I'll get rid of it.

PR #16, opened 2021-04-02
Add CI tests and CompatHelper

devmotion:
In addition to this PR, one has to add a SSH deploy key (e.g., by following the instructions here: https://github.com/JuliaRegistries/CompatHelper.jl#122-instructions-for-setting-up-the-ssh-deploy-key) to allow CompatHelper to trigger the CI tests.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -24,6 +24,11 @@ jobs:
           - windows-latest
         arch:
           - x64
+        include:
+          - version: '1'
+            os: ubuntu-latest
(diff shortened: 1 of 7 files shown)


phipsgabler:
LGTM as far as I can judge, but I'm not really familiar with GitHub workflows.  I have added the deploy key, can you check if it works correctly?

devmotion:
Unfortunately, I don't think it's possible to check it - we have to wait until CompatHelper runs or trigger it manually once the action is available in the default branch.

PR #17, opened 2021-04-05
Provide symbol in `VarName` constructor as type parameter

devmotion:
Constant propagation in the `VarName` does not work reliably as we noticed in https://github.com/TuringLang/DynamicPPL.jl/pull/221. Applying the changes in this PR improves performance noticeably.

IMO there is no reason to keep the existing type-unstable constructor. I deprecated it, therefore it is a non-breaking change.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,8 +3,7 @@ uuid = "7a57a42e-76ec-4ea3-a279-07e840d6d9cf"
 keywords = ["probablistic programming"]
 license = "MIT"
 desc = "Common interfaces for probabilistic programming"
-version = "0.1.2"
-
+version = "0.1.3"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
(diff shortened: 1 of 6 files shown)


phipsgabler:
I completely agree.  This is basically the same effect as the recommendation to use `Val{x}` instead of `Val(x)`, I believe.

marius311:
Following this deprecation, is there any succinct way to do `VarName.((:x,:y,:z))` without the warning?

phipsgabler:
Nothing shorter than `map(n -> VarName{n}(), (:x, :y, :z))`, I think. It's an inherently dynamic operation.

phipsgabler:
I mean, if this were a frequent practical problem, I can imagine a solution similar to `@gensym` where 

```julia
@varnames x y z[1]
```

expands (unhygienically) to 

```julia
x = VarName{:x}()
y = VarName{:y}()
var"z[1]" = @varname(z[1])
```

But unless there's a pressing need, I'll leave it as an exercise for the interested reader.

PR #18, opened 2021-04-07
Update badges and workflows

devmotion:
I completely missed that the package uses `main` and not `master` as default branch (BTW CompatHelper seems to work, it also noticed that a key COMPATHELPER_PRIV exists).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   push:
     branches:
-      - master
+      - main
   pull_request:
 
 jobs:
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -3,7 +3,7 @@ name: IntegrationTest
 on:
   push:
     branches:
-      - master
+      - main
   pull_request:
 
 jobs:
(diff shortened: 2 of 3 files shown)


phipsgabler:
Maybe I copied something wrongly.  I generated and added a new key pair now.

phipsgabler:
Is that enough of a check that this is good now: https://github.com/TuringLang/AbstractPPL.jl/runs/2303403984?check_suite_focus=true?  Or will the trigger work differently if there is actually something to update?

devmotion:
Unfortunately, it doesn't ensure that it's correct. The (possibly resolved) problem occurs only when CompatHelper actually opens a PR. Without the key or an incorrect key it can't trigger any other Github actions but this is not relevant for the check if any newer versions exist.

phipsgabler:
Hm, I thought we would notice it the next time CompatHelper runs, but appearently that is not so often.  Do you think I should just merge and wait, or wait for a test case, first?

devmotion:
Oh CompatHelper runs every day at midnight but the error only occurs (or hopefully not anymore...) if it opens a PR which doesn't happen too often since the package only depends on AbstractMCMC IIRC.

devmotion:
So I'd merge the PR without checking it - the worst thing that can happen is that it does not work the next time CompatHelper opens a PR and one has to close and reopen its PR manually to trigger the CI tests.

PR #22, opened 2021-07-10
Prettier show for `Colon`

torfjelde:
Currently:

```julia
julia> @varname(x[:,1])
x[Colon(),1]
```

On this branch:

```julia
julia> @varname(x[:,1])
x[:,1]
```

Seems like the latter is preferable, and will make a lot of downstream printing much nicer, e.g. chains obtained in Turing.

Only potential issue is that people might have code that access elements in a chain, etc. using `Colon()` in the string-rep, so maybe this should be a breaking release?

### DIFF
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -98,8 +98,8 @@ function Base.show(io::IO, vn::VarName)
     end
 end
 
-replace_colon_string(x) = x
-replace_colon_string(::Colon) = ":"
+prettify_index(x) = string(x)
+prettify_index(::Colon) = ":"
 
 """
     Symbol(vn::VarName)
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -93,11 +93,13 @@ function Base.show(io::IO, vn::VarName)
     print(io, getsym(vn))
     for indices in getindexing(vn)
         print(io, "[")
-        join(io, indices, ",")
+        join(io, map(replace_colon_string, indices), ",")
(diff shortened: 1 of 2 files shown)


yebai (approved):

torfjelde:
I'll make it a breaking release.

src/varname.jl
```diff
@@ -93,11 +93,13 @@ function Base.show(io::IO, vn::VarName)
...
     for indices in getindexing(vn)
         print(io, "[")
-        join(io, indices, ",")
+        join(io, map(replace_colon_string, indices), ",")
         print(io, "]")
     end
 end
 
+replace_colon_string(x) = x
+replace_colon_string(::Colon) = ":"
```
phipsgabler, line 102: Would it make sense to call this something like `prettyprint_index`, to allow for later extensibility?  E.g., I have thought about using negative indexing in my other PR, and then you'd want to have something like 
```julia
prettyprint_index(ix::InvertedIndex) = "-$(prettyprint_index(ix.skip))"
```

torfjelde: Yeah, I'm happy with that:)

torfjelde:
Eehh maybe we should require tests to pass before merging here too? The tests are now failing (I forgot to rename some stuff in the tests).

phipsgabler:
@torfjelde sounds like a good idea. Where can I activate that?

PR #24, opened 2021-07-13
Fix for bug introduced in #22

torfjelde:
Fixes bug introduced in #22


yebai:
Thanks for the fix, @torfjelde

PR #25, opened 2021-07-17
Fix begin/end replacement in varname expansion (fixes #23)

### LINKED ISSUE #23: Nested begin/end indexing no longer supported
Back in the day, doing something like `x[1][end]` was possible, but it no longer is:

```julia
julia> AbstractPPL.vinds(:(x[1][end]))
:((((lastindex)(var"##S#319"),),))
```

which references an undefined variable. This is because of the `Base.replace_ref_begin_end!`:

```julia
julia> Base.replace_ref_begin_end!(:(x[1][end]))
:(let var"##S#320" = x[1]
      var"##S#320"[(lastindex)(var"##S#320")]
  end)
```

which is then further altered afterwards in

https://github.com/TuringLang/AbstractPPL.jl/blob/34fc3e1ba12057993aa64fb850d392fa3b06a354/src/varname.jl#L357-L364

resulting in the incorrect generated code.

Is this intended?

phipsgabler:
The trick is to (1) first extract the indices, then replace the `begin`/`end`, and (2) make use of the internal `Base.replace_ref_(begin_)end_!`, which lets you specify the replacement yourself.

The latter seems a bit dubious, but I don't see another chance without a complete reimplementation of `Base.replace_ref_begin_end!`.  Any kind of recursive implementation will have that argument anyway, since you need to care about nested scopes, so it should be kinda stable... I hope.

I'd also like to include a test of the following kind:

```julia
julia> vinds(:(x[:, a[10][end], :]))
:(((:, let var"##S#266" = a[10]
              var"##S#266"[(lastindex)(var"##S#266")]
          end, :),))
```

But those gensyms won't be liked by the doctests.  Any idea how to nicely add it?

@torfjelde check it out.

### DIFF
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -224,10 +224,6 @@ A macro that returns an instance of [`VarName`](@ref) given a symbol or indexing
 The `sym` value is taken from the actual variable name, and the index values are put appropriately
 into the constructor (and resolved at runtime).
 
-NB: `begin` and `end` indexing can be used, but remember that they depend on _runtime values_ --
-their usage requires that the array over which the indexing expression is defined is defined, in
-order for `firstindex` and `lastindex` to work in the expanded code.
-
 ## Examples
 
 ```jldoctest
@@ -249,6 +245,13 @@ julia> @varname(x[1,2][1+5][45][3]).indexing
 julia> let a = [42]; @varname(a[1][end][3]); end
 a[1][1][3]
 ```
+
+!!! warning
+    As you can see in the last example, `@varname` does not do any bounds checking!
+
+    Only `begin` and `end` indexing, which depend on the _runtime size_ of the array, does that!
+    Their usage _requires_ that the array over which the indexing expression is defined, in
+    order for `firstindex` and `lastindex` to work in the expanded code.
 
 !!! compat "Julia 1.5"
     Using `begin` in an indexing expression to refer to the first index requires at least
@@ -401,11 +404,11 @@ function vinds(expr, head = vsym(expr))
         
         nixs = length(ixs)
         if nixs == 1
-            # for 1D indexing, just use `lastindex(x)`
+            # for linear indexing, we use `lastindex(x)`
             ixs[1], used = _replace_ref_begin_end(ixs[1], _index_replacement_for(S))
             used_S |= used
         elseif nixs > 1
-            # otherwise, we need `lastindex(x, i)`
+            # for cartesian indexing, we need `lastindex(x, i)`
             for i in eachindex(ixs)
                 ixs[i], used = _replace_ref_begin_end(ixs[i], _index_replacement_for(S, i))
                 used_S |= used
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -374,6 +374,9 @@ julia> x = [10 20]; eval(vinds(:(x[1, end])))
 julia> x = [[1, 2]]; eval(vinds(:(x[1][end])))
 ((1,), (2,))
 
+julia> x = ([1, 2], ); eval(vinds(:(x[1][end]))) # tuple
+((1,), (2,))
+
 julia> x = [fill([[10], [20, 30]], 2, 2, 2)]
        if VERSION < v"1.5.0-DEV.666"
             eval(vinds(:(x[1][2, end, :][2][end])))
@@ -389,7 +392,7 @@ function vinds(expr, head = vsym(expr))
     indexing = _straighten_indexing(expr)
     inds = Expr[]  # collection of result indices
     partial = head  # partial :ref expressions, used in caching
-    cached_exprs = Vector{Pair{Symbol, Expr}}()  # cache for partial expressions in a :let
+    cached_exprs = Vector{Pair{Symbol, Expr}}()  # cache for partial expressions going into a let
     
     for ixs in indexing
         # S becomes the name of the cached variable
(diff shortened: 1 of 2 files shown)


src/varname.jl
```diff
@@ -348,23 +363,30 @@ julia> vinds(:(x[end]))
 
 julia> vinds(:(x[1, end]))
 :(((1, (lastindex)(x, 2)),))
+
+julia> vinds(:(x[1][end]))
+:(((1,), ((lastindex)(x),)))
```
torfjelde, line 368: This is incorrect. You want `:(((1,), ((lastindex)(x, 2), )))`, right?

src/varname.jl
```diff
@@ -348,23 +363,30 @@ julia> vinds(:(x[end]))
...
+
+julia> vinds(:(x[1][2, end, :][3]))
+:(((1,), (2, (lastindex)(x), :), (3,)))
```
torfjelde, line 371: Same here

src/varname.jl
torfjelde: This is why I thought it would be difficult to fix btw.

phipsgabler: Ah :hankey: Maybe it's feasible to instead turn it around and replace the first head of a `:ref` within the `let`s instead.

phipsgabler:
I think it should work now.  But there's even more gensyms in the doctests, which it isn't gonna like.

torfjelde:
Just make them evaluated expression rather than the generated `Expr`, e.g.
```julia
julia> x = [[1.0, 2.0]]; eval(vinds(:(x[1][end])))
2.0
```

torfjelde:
Is this good for a review?

src/varname.jl
```diff
@@ -241,6 +245,9 @@ julia> @varname(x[:, 1][2]).indexing
 
 julia> @varname(x[1,2][1+5][45][3]).indexing
 ((1, 2), (6,), (45,), (3,))
+
+julia> let a = [42]; @varname(a[1][end][3]); end
```
phipsgabler, line 249: That's not really a meaningful example...

phipsgabler:
@torfjelde Yes, I think so.

src/varname.jl
```diff
@@ -334,37 +342,125 @@ macro vinds(expr::Union{Expr, Symbol})
...
+    cached_exprs = Vector{Pair{Symbol, Expr}}()  # cache for partial expressions in a :let
+    
+    for ixs in indexing
+        # S becomes the name of the cached variable
+        S = (partial == head) ? head : gensym(:S)
+        used_S = false
+        
+        nixs = length(ixs)
+        if nixs == 1
+            # for 1D indexing, just use `lastindex(x)`
```
torfjelde, line 401:
```suggestion
            # For linear indexing we use `lastindex(x)`.
```
I was a bit confused as to even if we're doing 1D-indexing, why don't just use use `_index_replacement_for(S, 1)`, but then I remembered there's such a thing as linear indexing:) So I think this is a bit clearer.

src/varname.jl
```diff
@@ -334,37 +342,125 @@ macro vinds(expr::Union{Expr, Symbol})
...
-function vinds(expr::Expr)
-    if Meta.isexpr(expr, :ref)
-        ex = copy(expr)
-        @static if VERSION < v"1.5.0-DEV.666"
-            Base.replace_ref_end!(ex)
+        if used_S && partial !== head
+            # cache that expression if we actually used it, and use the new name in the
+            # partial expression
+            push!(cached_exprs, S => partial)
+            partial = Expr(:ref, S, ixs...)
```
torfjelde, line 416: Should this be a `view` instead? This would avoid unnecessary allocations (at least on more recent versions of Julia).

EDIT: I guess one issue here is that it will exclude indexing of non-array-types, e.g. `Tuple`:
```julia
julia> x = ((rand(1, 2, 3), ), );

julia> eval(vinds(:(x[1][1][1,end,1])))
ERROR: MethodError: no method matching view(::Tuple{Tuple{Array{Float64, 3}}}, ::Int64)
Closest candidates are:
  view(::AbstractArray, ::Any...) where N at subarray.jl:174
```

So instead we should use `Base.maybeview`.

torfjelde (changes requested):
Good stuff! I've left a couple of comments, the most important one being that we should probably use `Base.maybeview` rathe than not.

IMO we should have integration tests for DPPL in AbstractPPL.jl, but for now, could you just run DPPL tests using this branch of APPL to verify that things work? In particular I'm curious about Julia 1.3.

Oh and could you bump the patch-version?

src/varname.jl
phipsgabler: I had tried a version with `view` here, but the main problem was that it returns 0-D views for scalar indices.  But I didn't know about `maybeview`, that's awesome!

phipsgabler:
> IMO we should have integration tests for DPPL in AbstractPPL.jl, but for now, could you just run DPPL tests using this branch of APPL to verify that things work? In particular I'm curious about Julia 1.3.

I had the same thought, but Hong recently told me they exist already: https://github.com/TuringLang/AbstractPPL.jl/blob/main/.github/workflows/IntegrationTest.yml.  Is that enough?

src/varname.jl
```diff
@@ -334,37 +342,125 @@ macro vinds(expr::Union{Expr, Symbol})
...
+    end
+end
+
+
+"""
+    _straighten_indexing(expr)
+
+Extract a list of lists of (raw) indices of an iterated `:ref` expression.
+
+```julia
```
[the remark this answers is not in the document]
phipsgabler: Does that work with a non-exported function?

torfjelde:
Oh yeah, crap. It's because we've bumped the minor version due to changes in `string` for `VarName`. We need to bump compat with DPPL. I'll get on that. Once that's done, the integration tests will run properly.

phipsgabler:
Closed in favour of #26, which implements a superset of this functionality.

PR #26, opened 2021-07-27
Using Setfield.jl's lenses to handle indexing

torfjelde:
Correctly parsing all the `VarName` stuff is _hard_, e.g. https://github.com/TuringLang/AbstractPPL.jl/pull/25, plus we're currently restricted to arrays only.

This PR allows us to defer all the indexing functionality to Setfield.jl's lenses! In addition, it allows us to potentially use `@set` and others within `@model` in DPPL to also support sampling elements of `NamedTuple`, etc.!

The only "major" change is the introduction of `concretize` which we'll have to use in `@model` to address runtime-based indexing, e.g. `x[end]`. But AFAIK this doesn't come at any greater disadvantages than the current impl.

This would address #23 and is further related to https://github.com/TuringLang/Turing.jl/issues/1668 and many other issues/ideas.

```julia
julia> x = (a = [1.0 2.0;], );

julia> vn = @varname(x.a[1, :])
x.a[1, :]

julia> AbstractPPL.concretize(vn, x)
x.a[1, :]

julia> vn = @varname(x.a[1, end][:]);

julia> AbstractPPL.concretize(vn, x)
x.a[1, 2][:]
```

EDIT: You can then do stuff like
```julia
julia> @model function demo(x, y)
           s ~ InverseGamma(2, 3)
           m ~ Normal(0, √s)
           for i in 1:length(x.a) - 1
               x.a[i] ~ Normal(m, √s)
           end

           x.a[end] ~ Normal(100.0, 1.0)
           y.a ~ Normal()
           
           return (; s, m, x, y)
       end
demo (generic function with 2 methods)

julia> struct MyCoolStruct
           a
       end

julia> m = demo(MyCoolStruct([1.0, ]), MyCoolStruct(missing));

julia> m()
(s = 6.097169879128847, m = -5.565352920139915, x = MyCoolStruct([1.0]), y = MyCoolStruct(-0.6953528936398823))

julia> m = demo(MyCoolStruct([1.0, ]), (a = missing, ));

julia> m()
(s = 0.7759718829619909, m = -0.5107335960569337, x = MyCoolStruct([1.0]), y = (a = 2.8120990475668273,))
```

Or we can be a bit mad:

```julia
julia> using DataFrames

julia> using Setfield: ConstructionBase

julia> function ConstructionBase.setproperties(df::DataFrame, patch::NamedTuple)
           # Only need `copy` because we'll replace entire columns
           columns = copy(DataFrames._columns(df))
           colindex = DataFrames.index(df)
           for k in keys(patch)
               columns[colindex[k]] = patch[k]
           end
           return DataFrame(columns, colindex)
       end

julia> @model function demo(x)
           s ~ InverseGamma(2, 3)
           m ~ Normal(0, √s)
           for i in 1:length(x.a) - 1
               x.a[i] ~ Normal(m, √s)
           end

           x.a[end] ~ Normal(100.0, 1.0)
           
           return x
       end
demo (generic function with 1 method)

julia> m = demo(df, (a = missing, ));

julia> m()
3×1 DataFrame
 Row │ a        
     │ Float64? 
─────┼──────────
   1 │   1.0
   2 │   2.0
   3 │  99.8838

julia> df
3×1 DataFrame
 Row │ a         
     │ Float64?  
─────┼───────────
   1 │       1.0
   2 │       2.0
   3 │ missing   

```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "7a57a42e-76ec-4ea3-a279-07e840d6d9cf"
 keywords = ["probablistic programming"]
 license = "MIT"
 desc = "Common interfaces for probabilistic programming"
-version = "0.2.0"
+version = "0.3.0"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
--- a/Project.toml
+++ b/Project.toml
@@ -7,6 +7,7 @@ version = "0.1.4"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
+Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
--- a/Project.toml
+++ b/Project.toml
@@ -8,7 +8,6 @@ version = "0.2.0"
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
-StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
 AbstractMCMC = "2, 3"
--- a/Project.toml
+++ b/Project.toml
@@ -11,4 +11,5 @@ Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
 
 [compat]
 AbstractMCMC = "2, 3"
+Setfield = "0.7.1"
 julia = "1"
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -106,7 +106,9 @@ function Base.show(io::IO, vn::VarName{<:Any, <:Lens})
     Setfield.print_application(io, vn.indexing)
 end
 
+# TODO: Should this really go here?
 Setfield.print_application(io::IO, l::IndexLens) = print(io, "[", join(map(prettify_index, l.indices), ", "), "]")
+Setfield.print_application(io::IO, l::DynamicIndexLens) = print(io, l, "(_)")
 
 prettify_index(x) = string(x)
 prettify_index(::Colon) = ":"
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -1,3 +1,5 @@
+using Setfield
+
 """
     VarName{sym}(indexing::Tuple=())
 
@@ -26,10 +28,10 @@ julia> @varname x[:, 1][1+1]
 x[Colon(),1][2]
 ```
 """
-struct VarName{sym, T<:Tuple}
+struct VarName{sym, T}
     indexing::T
 
-    VarName{sym}(indexing::Tuple=()) where {sym} = new{sym,typeof(indexing)}(indexing)
+    VarName{sym}(indexing=()) where {sym} = new{sym,typeof(indexing)}(indexing)
 end
 
 """
@@ -89,13 +91,18 @@ getindexing(vn::VarName) = vn.indexing
 Base.hash(vn::VarName, h::UInt) = hash((getsym(vn), getindexing(vn)), h)
 Base.:(==)(x::VarName, y::VarName) = getsym(x) == getsym(y) && getindexing(x) == getindexing(y)
 
-function Base.show(io::IO, vn::VarName)
+function Base.show(io::IO, vn::VarName{<:Any, <:Tuple})
     print(io, getsym(vn))
     for indices in getindexing(vn)
         print(io, "[")
         join(io, indices, ",")
         print(io, "]")
     end
+end
+
+function Base.show(io::IO, vn::VarName{<:Any, <:Lens})
+    print(io, getsym(vn))
+    Setfield.print_application(io, vn.indexing)
 end
 
 
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -493,34 +493,6 @@ function drop_escape(expr::Expr)
     return Expr(expr.head, map(x -> drop_escape(x), expr.args)...)
 end
 
-@static if VERSION ≥ v"1.5.0-DEV.666"
-    # TODO: Replace once https://github.com/jw3126/Setfield.jl/pull/155 has been merged.
-    function Setfield.lower_index(collection::Symbol, index, dim)
-        if Setfield.isexpr(index, :call)
-            return Expr(:call, Setfield.lower_index.(collection, index.args, dim)...)
-        elseif (index === :end)
-            if dim === nothing
-                return :($(Base.lastindex)($collection))
-            else
-                return :($(Base.lastindex)($collection, $dim))
-            end
-        elseif index === :begin
-            if dim === nothing
-                return :($(Base.firstindex)($collection))
-            else
-                return :($(Base.firstindex)($collection, $dim))
-            end
-        end
-        return index
-    end
-
-    function Setfield.need_dynamic_lens(ex)
-        return Setfield.foldtree(false, ex) do yes, x
-            (yes || x === :end || x === :begin || x === :_)
-        end
-    end
-end
-
 """
     @vsym(expr)
 
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -88,6 +88,24 @@ julia> getindexing(@varname(y))
 """
 getindexing(vn::VarName) = vn.indexing
 
+"""
+    get(obj, vn::VarName{sym})
+
+Alias for `get(obj, PropertyLens{sym}() ∘ vn.indexing)`.
+"""
+function Setfield.get(obj, vn::VarName{sym}) where {sym}
+    return Setfield.get(obj, PropertyLens{sym}() ∘ vn.indexing)
+end
+
+"""
+    set(obj, vn::VarName{sym}, value)
+
+Alias for `set(obj, PropertyLens{sym}() ∘ vn.indexing, value)`.
+"""
+function Setfield.set(obj, vn::VarName{sym}, value) where {sym}
+    return Setfield.set(obj, PropertyLens{sym}() ∘ vn.indexing, value)
+end
+
 
 Base.hash(vn::VarName, h::UInt) = hash((getsym(vn), getindexing(vn)), h)
 Base.:(==)(x::VarName, y::VarName) = getsym(x) == getsym(y) && getindexing(x) == getindexing(y)
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -1,5 +1,5 @@
 using Setfield
-import Setfield: PropertyLens, ComposedLens, IdentityLens, IndexLens
+import Setfield: PropertyLens, ComposedLens, IdentityLens, IndexLens, DynamicIndexLens
 
 """
     VarName{sym}(indexing::Tuple=())
@@ -48,7 +48,7 @@ julia> VarName(@varname(x[1][2:3]))
 x
 ```
 """
-function VarName(vn::VarName, indexing::Tuple = ())
+function VarName(vn::VarName, indexing = ())
     return VarName{getsym(vn)}(indexing)
 end
 
@@ -105,6 +105,8 @@ function Base.show(io::IO, vn::VarName{<:Any, <:Lens})
     print(io, getsym(vn))
     Setfield.print_application(io, vn.indexing)
 end
+
+Setfield.print_application(io::IO, l::IndexLens) = print(io, "[", join(map(prettify_index, l.indices), ", "), "]")
 
 prettify_index(x) = string(x)
 prettify_index(::Colon) = ":"
@@ -247,7 +249,24 @@ subsumes(t::PropertyLens, u::PropertyLens) = false
 #        (but neither did old implementation).
 subsumes(t::IndexLens, u::IndexLens) = _issubindex(t.indices, u.indices)
 
+"""
+    concretize(l::Lens, x)
 
+Return `l` instantiated on `x`, i.e. any runtime information evaluated using `x`.
+"""
+
+concretize(I::Lens, x) = I
+concretize(I::DynamicIndexLens, x) = IndexLens(I.f(x))
+function concretize(I::ComposedLens, x)
+    x_inner = get(x, I.outer)
+    return ComposedLens(concretize(I.outer, x), concretize(I.inner, x_inner))
+end
+"""
+    concretize(vn::VarName, x)
+
+Return `vn` instantiated on `x`, i.e. any runtime information evaluated using `x`.
+"""
+concretize(vn::VarName, x) = VarName(vn, concretize(vn.indexing, x))
 
 """
     @varname(expr)
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -382,21 +382,35 @@ end
 varname(sym::Symbol, concretize=false) = :($(AbstractPPL.VarName){$(QuoteNode(sym))}())
 function varname(expr::Expr, concretize=false)
     if Meta.isexpr(expr, :ref) || Meta.isexpr(expr, :.)
-        sym = vsym(expr)
-        # Convert `expr` into something `lensmacro` can parse.
-        inds = Setfield.lensmacro(identity, replace_basesym(expr, :_))
+        # Split into object/base symbol and lens.
+        sym_escaped, lens = Setfield.parse_obj_lens(expr)
+        # Setfield.jl escapes the return symbol, so we need to unescape
+        # to call `QuoteNode` on it.
+        sym = drop_escape(sym_escaped)
 
-        # TODO: Can we do better, i.e. only check if we have `DynamicLens`?
-        return if concretize
-            :($(AbstractPPL.concretize)($(AbstractPPL.VarName){$(QuoteNode(sym))}($inds), $(esc(sym))))
+        return if Setfield.need_dynamic_lens(expr)
+            :($(AbstractPPL.concretize)($(AbstractPPL.VarName){$(QuoteNode(sym))}($lens), $sym_escaped))
         else
-            :($(AbstractPPL.VarName){$(QuoteNode(sym))}($inds))
+            :($(AbstractPPL.VarName){$(QuoteNode(sym))}($lens))
         end
     else
         error("Malformed variable name $(expr)!")
     end
 end
 
+drop_escape(x) = x
+function drop_escape(expr::Expr)
+    Meta.isexpr(expr, :escape) && return drop_escape(expr.args[1])
+    return Expr(expr.head, map(x -> drop_escape(x), expr.args)...)
+end
+
+@static if VERSION ≥ v"1.5.0-DEV.666"
+    function Setfield.need_dynamic_lens(ex)
+        return Setfield.foldtree(false, ex) do yes, x
+            yes || x === :end || x === :begin || x === :_
+        end
+    end
+end
 
 """
     replace_basesym(expr, sub::Symbol)
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -336,9 +336,12 @@ x.a[1,2][:]
 concretize(vn::VarName, x) = VarName(vn, concretize(vn.indexing, x))
 
 """
-    @varname(expr)
+    @varname(expr[, concretize])
 
 A macro that returns an instance of [`VarName`](@ref) given a symbol or indexing expression `expr`.
+
+If `concretize` is `true`, the resulting expression will be wrapped in a [`concretize`](@ref) call.
+This is useful if you for example want to ensure that no `Setfield.DynamicLens` is used.
 
 The `sym` value is taken from the actual variable name, and the index values are put appropriately
 into the constructor (and resolved at runtime).
@@ -367,7 +370,7 @@ julia> @varname(x[1,2][1+5][45][3]).indexing
     Julia 1.5.
 """
 macro varname(expr::Union{Expr, Symbol}, concretize=false)
-    return varname(expr, concretize=concretize)
+    return varname(expr, concretize)
 end
 
 varname(sym::Symbol, concretize=false) = :($(AbstractPPL.VarName){$(QuoteNode(sym))}())
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -20,13 +20,13 @@ indexing expression through the [`@varname`](@ref) convenience macro.
 
 ```jldoctest
 julia> vn = VarName{:x}(((Colon(), 1), (2,)))
-x[:, 1][2]
+x[:,1][2]
 
 julia> vn.indexing
 ((Colon(), 1), (2,))
 
 julia> @varname x[:, 1][1+1]
-x[:, 1][2]
+x[:,1][2]
 ```
 """
 struct VarName{sym, T}
@@ -80,7 +80,7 @@ Return the indexing tuple of the Julia variable used to generate `vn`.
 
 ```jldoctest
 julia> getindexing(@varname(x[1][2:3]))
-((1,), (2:3,))
+(@lens _[1][2:3])
 
 julia> getindexing(@varname(y))
 ()
@@ -326,15 +326,15 @@ Return `vn` instantiated on `x`, i.e. any runtime information evaluated using `x
 julia> x = (a = [1.0 2.0;], );
 
 julia> vn = @varname(x.a[1, :])
-x.a[1, :]
+x.a[1,:]
 
 julia> AbstractPPL.concretize(vn, x)
-x.a[1, :]
+x.a[1,:]
 
 julia> vn = @varname(x.a[1, end][:]);
 
 julia> AbstractPPL.concretize(vn, x)
-x.a[1, 2][:]
+x.a[1,2][:]
 ```
 """
 concretize(vn::VarName, x) = VarName(vn, concretize(vn.indexing, x))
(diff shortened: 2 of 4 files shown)


src/varname.jl
```diff
@@ -98,6 +101,15 @@ function Base.show(io::IO, vn::VarName)
     end
 end
 
+function Base.show(io::IO, vn::VarName{<:Any, <:Lens})
+    print(io, getsym(vn))
+    Setfield.print_application(io, vn.indexing)
+end
+
+# TODO: Should this really go here?
```
torfjelde, line 109: The answer is no of course, but I'm not bothered to start impl this in a nicer way until we've decided that this is a path we want to take.

src/varname.jl
```diff
@@ -294,7 +377,7 @@ function vsym end
 
 vsym(expr::Symbol) = expr
 function vsym(expr::Expr)
-    if Meta.isexpr(expr, :ref)
+    if Meta.isexpr(expr, :ref) || Meta.isexpr(expr, :.)
```
phipsgabler, line 380: We might also need to handle `GlobalRef` here.

phipsgabler:
Wouldn't it be simpler to only allow "concrete" `VarName`s, and directly expand `@varname` to something like 

```julia
let dyn_vn = $(<dynamic varname as currently written>)
    concretize(dyn_vn, $(<the vsym>))
end
```
?

torfjelde:
> Wouldn't it be simpler to only allow "concrete" VarNames, and directly expand @varname to something like

I've added this now:) The only thing we needed to be careful was that the symbol might not be defined yet.

src/varname.jl
```diff
@@ -231,37 +359,76 @@ julia> @varname(x).indexing
...
+
+julia> AbstractPPL.replace_basesym(:(x), :_)
+:x
+
+julia> AbstractPPL.replace_basesym(:(1), :_)
+1
+```
+
+"""
+replace_basesym(x, sub::Symbol) = x
```
phipsgabler, line 420: https://github.com/jw3126/Setfield.jl/blob/544346fce31eb904a951eb651e69fe0fe24490b2/src/sugar.jl#L69

Also, perhaps we can reuse [parse_obj_lens](https://github.com/jw3126/Setfield.jl/blob/544346fce31eb904a951eb651e69fe0fe24490b2/src/sugar.jl#L133).

src/varname.jl
```diff
@@ -231,37 +359,76 @@ julia> @varname(x).indexing
...
+(@lens _[1, 2][6][45][3])
 ```
 
 !!! compat "Julia 1.5"
     Using `begin` in an indexing expression to refer to the first index requires at least
     Julia 1.5.
 """
-macro varname(expr::Union{Expr, Symbol})
-    return esc(varname(expr))
+macro varname(expr::Union{Expr, Symbol}, concretize=false)
```
phipsgabler, line 378: This really needs to be `concretize::Bool=false`, otherwise the `if` will fail if you pass an arbitrary expression.  (Or you splice the expression in a returned `if`, but I'd prefer the check to happen at expansion time.)

Project.toml
```diff
@@ -7,6 +7,8 @@ version = "0.2.0"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
+Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
+StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
```
phipsgabler, line 11: Not needed.

src/varname.jl
```diff
@@ -98,6 +106,21 @@ function Base.show(io::IO, vn::VarName)
     end
 end
 
+function Base.show(io::IO, vn::VarName{<:Any, <:Lens})
```
phipsgabler, line 109: There's something off with `DynamicIndexLens`:

```julia
julia> @varname x[1][end]
x[1]Setfield.DynamicIndexLens{var"#3#4"}(var"#3#4"())(_)
```

src/varname.jl
```diff
@@ -89,7 +92,12 @@ getindexing(vn::VarName) = vn.indexing
 Base.hash(vn::VarName, h::UInt) = hash((getsym(vn), getindexing(vn)), h)
 Base.:(==)(x::VarName, y::VarName) = getsym(x) == getsym(y) && getindexing(x) == getindexing(y)
```
phipsgabler, line 93: We need to be careful here now... what about `DynamicIndexLens`, containing a closure?

src/varname.jl
```diff
@@ -26,10 +29,10 @@ julia> @varname x[:, 1][1+1]
 x[:,1][2]
 ```
 """
-struct VarName{sym, T<:Tuple}
+struct VarName{sym, T}
     indexing::T
 
-    VarName{sym}(indexing::Tuple=()) where {sym} = new{sym,typeof(indexing)}(indexing)
+    VarName{sym}(indexing=()) where {sym} = new{sym,typeof(indexing)}(indexing)
```
phipsgabler, line 35: We could entirely get rid of tuples and use `IdentityLens` instead.

(And then require `T<:Lens` or so.)

src/varname.jl
phipsgabler: Maybe the more stable option is to use `parse_obj_lens` first to split the `vsym` and the indexing, and then just operate on the constructed VarName?  Now that the `vinds` is finally going to be removed from the tilde functions, is there any more need for `vsym`/`vinds` on their own?

phipsgabler (changes requested):
Two more things that I find should be done:

- Move over the updated docstrings and doctests from [my other PR](https://github.com/TuringLang/AbstractPPL.jl/pull/25/files), especially wrt. "dynamic" (`begin`/`end`) stuff.
- Implement (and reexport) `Setfield.get(obj, ::VarName)`, to replace the common pattern
  ```julia
  ixs = getindexing(vn)
  val = _getindex(x, ixs)
  ```

And if we're really serious about `begin`, let's fix https://github.com/jw3126/Setfield.jl/issues/154 ourselves.

src/varname.jl
torfjelde: That's just my weird show impl below since Setfield.jl doesn't have a `print_application` (which is what we want here) for `DynamicIndexLens`. I didn't give it too much thought since we're not going to be using `DynamicLens` in DPPL anyways, i.e. we will always concretize.

src/varname.jl
torfjelde: `replace_underscore` is the opposite of what we want here:) And we don't really want to do something like `postwalk` since we're _only_ interested in replacing the base symbol.

torfjelde: `parse_obj_lens` is a great call though!

src/varname.jl
```diff
@@ -231,37 +359,76 @@ julia> @varname(x).indexing
...
-        sym, inds = vsym(expr), vinds(expr)
-        return :($(AbstractPPL.VarName){$(QuoteNode(sym))}($inds))
+varname(sym::Symbol, concretize=false) = :($(AbstractPPL.VarName){$(QuoteNode(sym))}())
+function varname(expr::Expr, concretize=false)
+    if Meta.isexpr(expr, :ref) || Meta.isexpr(expr, :.)
+        sym = vsym(expr)
+        # Convert `expr` into something `lensmacro` can parse.
+        inds = Setfield.lensmacro(identity, replace_basesym(expr, :_))
+
+        # TODO: Can we do better, i.e. only check if we have `DynamicLens`?
```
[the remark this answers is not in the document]
torfjelde: Ooooh nice!

src/varname.jl
torfjelde: This is exactly why I made it `false` :+1:

phipsgabler:
Something else that just came to my mind: do we need a convenience function to derive the "shape" a varname describes?  With something like 

```julia
 Setfield.ComposedLens{Setfield.IndexLens{Tuple{Int64}}, Setfield.IndexLens{Tuple{Int64}}}}
```

we'd previously have known that this indexes of the form `[.][.]` by just inspecting the tuple types (or `length`) of the indexing, `((i,), (j,))`.  The same information _is_ still preserved in the `ComposedLens`, but more hacky to get out.

src/varname.jl
torfjelde: I agree with the usage of `parse_obj_lens`; that will simplify things:)

Regarding `vsym`/`vinds`, I don't know :confused: IMO `vsym` can be nice still, as it's just a nice way of extracting the symbol used as a reference, but `vinds` I agree is useless now.

src/varname.jl
torfjelde: In that case they won't be equal, right? Which IMO seems like sensible behavior, no? They're not the same but they have the same behavior. 

Seems like the only other alternative is to completely drop the definition of `==` which also seems a bit much?

phipsgabler:
We should also contemplate how completely we should embrace (and communicate, and export) the fact that `VarNames` are then, in many respects, lenses.  Like, there's alreay the composition method, and we should have `get` and `set` working, too.  

Then, the only addition we make is that by the additional name, the varnames can be used as dictionary keys for parts of structures.  From some point of view, this is really a special "VsymLens" composed with the indexing lens, if we view it as a lens for VarInfo (the "VsymLens" can be neither a `PropertyLens`, nor an `IndexLens` with symbol key, because we don't know beforehand how a VarInfo is implemented).  On the other hand, this perspective does not work so well when applied to variables (`get(x, @varname(x[1]))`).

Also, it might be practical to have more typing support for concreteness, since only concrete lenses make sense as dictionary keys.  One option could be to include a `true`/`false` type tag, but I don't really like that.  Two subtypes would be nicer, but complicate things too much.  Maybe it suffices to have an `@generatated` function `isconcrete`?  (Or we don't check it at all -- I can't say how much need there really is.)

(What again are the use cases for dynamic varnames?)

src/varname.jl
phipsgabler: Yeah but the way you wrote it, withouth the type bound, it will still allow other expressions, which are not boolean literals, and then fail with a `TypeError` somwhere inside, instead of a `MethodError` directly.

src/varname.jl
torfjelde: Not opposed to doing that. I was only slightly worried that that this would have some consequences downstream and was thinking that we could support both. Buuuut it's probably a good idea to just tear that bandaid off immediately :shrug:

torfjelde:
> We should also contemplate how completely we should embrace (and communicate, and export) the fact that VarNames are then, in many respects, lenses. Like, there's alreay the composition method, and we should have get and set working, too.

I see what you're saying here and I agree that maybe it would make sense to fully embrace, though, as you point out, it's unclear how it fits or rather how useful it will be to make it a lense if it's only going to be used in a possibly counter-intuitive way on `VarInfo` (as you said, it's neither an `IndexLens` nor a `PropertyLens`).

My stance on this: let's leave that discussion for future work rather than this PR :shrug: There's already a loooot of utility in this as it is. But agree, interesting things to think about!

> (What again are the use cases for dynamic varnames?)

Well, we need to construct them before we `concretize` them:) But other than that, since we _can_ support it without any cost (to the current approach), why not?

torfjelde:
> Something else that just came to my mind: do we need a convenience function to derive the "shape" a varname describes? With something like
> 
> ```julia
>  Setfield.ComposedLens{Setfield.IndexLens{Tuple{Int64}}, Setfield.IndexLens{Tuple{Int64}}}}
> ```
> 
> we'd previously have known that this indexes of the form `[.][.]` by just inspecting the tuple types (or `length`) of the indexing, `((i,), (j,))`. The same information _is_ still preserved in the `ComposedLens`, but more hacky to get out.

This is effectively what I do in the `combine_indices` method in this PR (because I'm too lazy to reimplement the `_issubindex` / want to ensure that we _at least_ preserve the current behavior), no?

src/varname.jl
phipsgabler: Agree to both, this was more like... warn the _callers_ of that, and check that we ourselves don't fall over this (which should be easy, though, as we are only concerned with concretized varnames anyway).

phipsgabler:
> My stance on this: let's leave that discussion for future work rather than this PR :shrug: There's already a loooot of utility in this as it is. But agree, interesting things to think about!

Yeah, sure -- again I'm only recording things for discussion :D I'll copy the points over to the discussion section then (just thought that the thread there is a bit stalled now).

> This is effectively what I do in the `combine_indices` method in this PR (because I'm too lazy to reimplement the `_issubindex` / want to ensure that we _at least_ preserve the current behavior), no?

Ah, cool, I didn't think very much about that function.

torfjelde:
> Hm. I don't disagree to that reasoning; but I was coming from the perspective that if you can, you should not even allow the construction of invalid objects. And dynamic lenses as varinfo parts are very much invalid. But we might very well just check that or let them error.

Putting some checks in for `VarInfo` might be a good idea :+1:

src/varname.jl
torfjelde: Ah I see what you're saying, haha. I completely misparsed it earlier.

src/varname.jl
```diff
@@ -26,29 +29,42 @@ julia> @varname x[:, 1][1+1]
...
-    return VarName{getsym(vn)}(indexing)
+VarName(vn::VarName, indexing::Lens=IdentityLens()) = VarName{getsym(vn)}(indexing)
+
+function VarName(vn::VarName, indexing::Tuple)
+    return VarName{getsym(vn)}(tuple2indexlens(indexing))
 end
 
+tuple2indexlens(indexing::Tuple{}) = IdentityLens()
+tuple2indexlens(indexing::Tuple{<:Tuple}) = IndexLens(first(indexing))
+tuple2indexlens(indexing::Tuple) = IndexLens(first(indexing)) ∘ tuple2indexlens(indexing[2:end])
```
phipsgabler, line 67:
```suggestion
tuple2indexlens(indexing::Tuple) = mapfoldr(IndexLens, ∘, indexing, init=IdentityLens())
```

src/varname.jl
```diff
@@ -26,29 +29,42 @@ julia> @varname x[:, 1][1+1]
...
-struct VarName{sym, T<:Tuple}
+struct VarName{sym, T<:Lens}
     indexing::T
 
-    VarName{sym}(indexing::Tuple=()) where {sym} = new{sym,typeof(indexing)}(indexing)
+    VarName{sym}(indexing=IdentityLens()) where {sym} = new{sym,typeof(indexing)}(indexing)
 end
 
+# A bit of backwards compatibility.
+# TODO: Should we deprecate this?
```
phipsgabler, line 39: I'm not unhappy with keeping around that variant.

src/varname.jl
phipsgabler: Also, `2indexlens` -- isn't that a bit misleading, as the function does not return an index lens except in one special case (or never, when implemented as a fold)?  Maybe `tuple2indexlenses`?

phipsgabler: (OK, `compose` is actualy [doing the intelligent thing](https://github.com/jw3126/Setfield.jl/blob/3ce508dfaf680e7aa31997b61568a75913a69ea7/src/lens.jl#L132) and discards `IdentityLens`.)

src/varname.jl
```diff
@@ -77,26 +93,55 @@ Return the indexing tuple of the Julia variable used to generate `vn`.
...
+
 
 Base.hash(vn::VarName, h::UInt) = hash((getsym(vn), getindexing(vn)), h)
 Base.:(==)(x::VarName, y::VarName) = getsym(x) == getsym(y) && getindexing(x) == getindexing(y)
 
-function Base.show(io::IO, vn::VarName)
+# Composition rules similar to the standard one for lenses, but we need a special
+# one for the "empty" `VarName{..., Tuple{}}`.
+Base.:∘(vn::VarName{sym,<:IdentityLens}, lens::Lens) where {sym} = VarName{sym}(lens)
+Base.:∘(vn::VarName{sym,<:Lens}, lens::Lens) where {sym} = VarName{sym}(vn.indexing ∘ lens)
```
phipsgabler, line 129: The comment is obsolete now; and [`compose` specializes on `IdentityLens`](https://github.com/jw3126/Setfield.jl/blob/3ce508dfaf680e7aa31997b61568a75913a69ea7/src/lens.jl#L132).  Additionally, shouldn't we rather add a method to `Setfield.compose` directly?

```suggestion
Setfield.compose(vn::VarName{sym,<:Lens}, lens::Lens) where {sym} = VarName{sym}(vn.indexing ∘ lens)
```

src/varname.jl
```diff
@@ -77,26 +93,55 @@ Return the indexing tuple of the Julia variable used to generate `vn`.
...
 julia> getindexing(@varname(y))
 ()
 ```
 """
 getindexing(vn::VarName) = vn.indexing
 
+"""
+    get(obj, vn::VarName{sym})
+
+Alias for `get(obj, PropertyLens{sym}() ∘ vn.indexing)`.
```
phipsgabler, line 107: This will work with `SimpleVarInfo`, but not untyped `VarInfo`, and I'm not sure about typed `VarInfo`.  My original idea was to have `get(obj, vn) = get(obj, vn.indexing)`, but on a second view, this doesn't immediately seem better or worse than this variant.  Some combination will always be confusing... 

Maybe we could document the intent as: this getter is only thought for internal usage, whereas the pattern for `VarInfo` access is always with the varname as a _key_, so `getindex(vi, vn)`, which can then, if suitable, dispatch to `get(vi, vn)` or `get(vi[sym], vn.indexing)` or whatever.

src/varname.jl
```diff
@@ -211,16 +379,56 @@ const ConcreteIndex = Union{Int, AbstractVector{Int}} # this include all kinds o
...
+concretize(vn::VarName, x) = VarName(vn, concretize(vn.indexing, x))
 
 """
-    @varname(expr)
+    @varname(expr[, concretize])
 
 A macro that returns an instance of [`VarName`](@ref) given a symbol or indexing expression `expr`.
 
+If `concretize` is `true`, the resulting expression will be wrapped in a [`concretize`](@ref) call.
+This is useful if you for example want to ensure that no `Setfield.DynamicLens` is used.
```
phipsgabler, line 430:
```suggestion
This is useful if you for example want to ensure that no `Setfield.DynamicLens` is used, and all usages of `begin` and `end` are resolved immediately from the scope of the expanded expression.
```

Something like that, and perhaps add an the example usage from the `@varname` doctest here, too.

src/varname.jl
torfjelde: Maybe `indextuple2lens`?

src/varname.jl
torfjelde: Sure, this seems like a good idea :+1:

And the `IdentityLens` dispatch was really just there for when we could have either `Lens` or `Tuple`, not to try and "optimize" away anything (I didnt' check, but assumed that it indeed specialized on `IdentityLens`). But this is obsolete now:)

src/varname.jl
torfjelde: But forever or temporarily? If forever (it's honestly a convenient constructor for `VarName` dealing with arrays only), then there's no need to deprecate. If temporarily we should deprecate.

torfjelde: I'm fine with either.

src/varname.jl
torfjelde: Btw, reason why I don't use `mapreduce` here is just because I'm more confident that the compiler can handle the recursion rather than in `mapreduce`. I've found `mapreduce` to be a bit unreliable when it comes to type inference, e.g. the `find_indexing` or whatever it's called in DynamicPPL that I discovered a couple of weeks ago.

The renaming I'm down for though. Happy with `indextuple2lens`?

src/varname.jl
torfjelde: Ehhh it won't even work for `SimpleVarInfo`. `SimpleVarInfo` has a field that is a `NamedTuple` on which this would work, i.e. `svi.θ.x`.

We also use `VarName` quite a bit to index into fields of contexts, etc., e.g. in `ConditionContext` we'd use `get(context.values, vn)`, which is why I made this choice. If we want special behavior when used on `VarInfo` or `SimpleVarInfo`, this should just be done through different methods rather than overloading `get`. Which is I guess exactly what you're saying at the end there regarding `getindex(vi, vn)` :)

> Maybe we could document the intent as: this getter is only thought for internal usage

I don't see a reason why this shouldn't be used externally :confused: It's just an alias to some functionality, as it says above, that's it. We're not selling `VarName` as an implementation of a `Lens` or anything, it's just a convenient overload that applies to most of the use-cases.

src/varname.jl
torfjelde: Sure :+1: How should do this though? Warning will be slow. Error is too much. Docstring?

torfjelde:
@phipsgabler Maybe we should disallow the use of `DynamicLens` in `VarName`, at least for now? 

I was trying to "fix" `show` and realized that it can have quite unfortunate consequences if a `DynamicLens` somehow slips through, for example you could end up with something like `Symbol("VarName{:x}((@lens _[1]) ∘ Setfield.DynamicIndexLens{var\"#25#26\"}(var\"#25#26\"()))")` in your chain, which is, well, undesirable.

So, maybe we do the opposite of what we thought originally: restrict for now, and then we can think about allowing it later?

src/varname.jl
torfjelde: Actually we need to overload `\circ` not `Setfield.compose` because `Setfield.compose` will only dispatch if `\circ` is called with lenses as arguments.

src/varname.jl
```diff
@@ -26,29 +29,65 @@ julia> @varname x[:, 1][1+1]
...
 ```
 """
-struct VarName{sym, T<:Tuple}
+struct VarName{sym,T<:Lens}
     indexing::T
 
-    VarName{sym}(indexing::Tuple=()) where {sym} = new{sym,typeof(indexing)}(indexing)
+    function VarName{sym}(indexing=IdentityLens()) where {sym}
+        # TODO: Should we completely disallow or just `@warn`?
+        # TODO: Does this affect performance?
```
torfjelde, line 37: From DPPL benchmarks it doesn't seem to matter (this should be compiled away anways).

src/varname.jl
phipsgabler: > it's honestly a convenient constructor for `VarName` dealing with arrays only

Exactly my reasoning.

src/varname.jl
```diff
@@ -26,29 +29,64 @@ julia> @varname x[:, 1][1+1]
 x[:,1][2]
 ```
 """
-struct VarName{sym, T<:Tuple}
+struct VarName{sym,T<:Lens}
     indexing::T
```
torfjelde, line 33: Should we rename `indexing` to `lens` @phipsgabler ?

src/varname.jl
```diff
@@ -77,27 +115,62 @@ Return the indexing tuple of the Julia variable used to generate `vn`.
...
+_print_application(io::IO, l::Lens) = Setfield.print_application(io, l)
+function _print_application(io::IO, l::ComposedLens)
+    _print_application(io, l.outer)
+    _print_application(io, l.inner)
+end
+_print_application(io::IO, l::IndexLens) =
+    print(io, "[", join(map(prettify_index, l.indices), ","), "]")
+# This is a bit weird but whatever. We're almost always going to
+# `concretize` anyways.
+_print_application(io::IO, l::DynamicIndexLens) = print(io, l, "(_)")
```
torfjelde, line 172: Should we just remove this since we're going to warn (and possibly error) if we try to construct `VarName` using a `DynamicIndexLens`?

src/varname.jl
```diff
@@ -26,29 +29,64 @@ julia> @varname x[:, 1][1+1]
 x[:,1][2]
 ```
 """
-struct VarName{sym, T<:Tuple}
+struct VarName{sym,T<:Lens}
     indexing::T
 
-    VarName{sym}(indexing::Tuple=()) where {sym} = new{sym,typeof(indexing)}(indexing)
+    function VarName{sym}(indexing=IdentityLens()) where {sym}
+        # TODO: Should we completely disallow or just `@warn` of limited support?
```
torfjelde, line 36: What do you think @phipsgabler ?

torfjelde:
AFAIK the only part that is left is bumping the minor version. Tests (except IntegrationTest, of course) are passing + pass in the corresponding DPPL PR.

Oh btw, @devmotion , do you have any thoughts on this before we make the final call?

src/varname.jl
phipsgabler: I wanted to suggest that already, but expected someone to complain about incompatibility!  I'm all for it.

src/varname.jl
phipsgabler: Not sure... I think we should allow dynamic representation for generality, but also have a means of "checked construction".  Kind of like the `Symmetric` etc. wrappers in linear algebra work.  

What about giving that functionality to `concretize`, letting it fail for dynamic lenses?  Then we the results of `concretize` and `@varname` are guaranteed to be conrete (or fail), but you can still use the `VarName` constructor directly?

phipsgabler:
> AFAIK the only part that is left is bumping the minor version.

Are we *sure* the benchmark regression in DPPL is not caused by some oversight with the lens system, and not how it is used in the model macro?  Because if not, I'd rather find out and fix that first...

src/varname.jl
```diff
@@ -194,73 +271,280 @@ function subsumes(u::VarName, v::VarName)
...
+# us that it SHOULD be correct. I'll leave it as is for now to ensure that
+# we preserve the status quo, but I'm confused.
 _issubrange(i::Colon, j::ConcreteIndex) = true
 
+"""
+    concretize(l::Lens, x)
 
+Return `l` instantiated on `x`, i.e. any runtime information evaluated using `x`.
+"""
+
```
devmotion, line 427: This empty line seems wrong, I don't think the docstring will show up anywhere.

torfjelde:
This PR was not ready to merged.

PR #29, opened 2021-08-20
Fix docstring

devmotion:
Addresses https://github.com/TuringLang/AbstractPPL.jl/pull/26#discussion_r693001923.


torfjelde (approved):
Thanks!

PR #31, opened 2021-09-09
Clean-up for new release

torfjelde:
There was some minor left-over work from the lens-PR. This PR does that work. AFAIK we should be good for a release after this.

### DIFF
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -2,19 +2,18 @@ using Setfield
 using Setfield: PropertyLens, ComposedLens, IdentityLens, IndexLens, DynamicIndexLens
 
 """
-    VarName{sym}(indexing::Tuple=())
+    VarName{sym}(lens::Lens=IdentityLens())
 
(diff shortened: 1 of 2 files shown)


src/varname.jl
```diff
@@ -2,42 +2,41 @@ using Setfield
...
 x[:,1][2]
 ```
 """
 struct VarName{sym,T<:Lens}
-    indexing::T
+    lens::T
 
-    function VarName{sym}(indexing=IdentityLens()) where {sym}
+    function VarName{sym}(lens=IdentityLens()) where {sym}
         # TODO: Should we completely disallow or just `@warn` of limited support?
```
torfjelde, line 35: IMO we disallow for now, and then we can consider enabling dynamic lenses if we have a good use-case for it.

yebai: Sounds good.

yebai (approved):

PR #41, opened 2021-11-08
Adopt DensityInterface.jl

phipsgabler:
This is in the hope of finally having a shared logdensity function across "probabilistic packages". Distributions.jl is already pretty much on board, Soss.jl hopefully will join later as soon as questions about how to integrate this with properly done measure theory (implicit/explicit base measures) have been resolved.

The consensus was to let the name be `logdensityof` (and `densityof`) to avoid name clashes. The interface is intended for both types that "have a density" and "are a density". 

The other part of DensityInterface.jl is a trait, `hasdensity`, which for `AbstractProbabilisticModel` I have defaulted to `true`. I think likelihood-free inference is out of scope for the kind of models we target -- is that a reasonable assumption?

See https://github.com/JuliaStats/Distributions.jl/pull/1416, https://github.com/JuliaMath/DensityInterface.jl/issues/1 and https://github.com/JuliaMath/DensityInterface.jl/issues/3.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -12,6 +12,6 @@ Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
 
 [compat]
 AbstractMCMC = "2, 3"
-DensityInterface = "0.3"
+DensityInterface = "0.4"
 Setfield = "0.7.1, 0.8"
 julia = "1"
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "7a57a42e-76ec-4ea3-a279-07e840d6d9cf"
 keywords = ["probablistic programming"]
 license = "MIT"
 desc = "Common interfaces for probabilistic programming"
-version = "0.3.1"
+version = "0.4"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
--- a/Project.toml
+++ b/Project.toml
@@ -7,9 +7,11 @@ version = "0.3.1"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
+DensityInterface = "b429d917-457f-4dbc-8f4c-0cc954292b1d"
 Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
 
 [compat]
 AbstractMCMC = "2, 3"
+DensityInterface = "0.3"
 Setfield = "0.7.1, 0.8"
 julia = "1"
--- a/interface.md
+++ b/interface.md
@@ -30,7 +30,7 @@ functions
 - `condition(::Model, ::Trace) -> ConditionedModel`
 - `decondition(::ConditionedModel) -> GenerativeModel`
 - `sample(::Model, ::Sampler = Exact(), [Int])` (from `AbstractMCMC.sample`)
-- `logdensityof(::Model, ::Trace)` and `density(::Model, ::Trace)` (from
+- `logdensityof(::Model, ::Trace)` and `densityof(::Model, ::Trace)` (from
   [DensityInterface.jl](https://github.com/JuliaMath/DensityInterface.jl))
 
 
@@ -215,10 +215,11 @@ the concrete model instance.
 
 ```
 DensityInterface.densityof(d, x) = exp(logdensityof(d, x))
-DensityInterface.logdensity(d) = Base.Fix1(logdensity, d)
+DensityInterface.logdensityof(d) = Base.Fix1(logdensityof, d)
+DensityInterface.densityof(d) = Base.Fix1(densityof, d)
 ```
 
-are provided automatically. 
+are provided automatically (repeated here for clarity). 
 
 Note that `logdensityof` strictly generalizes `logpdf`, since the posterior density will of course
 in general be unnormalized and hence not a probability density.
--- a/interface.md
+++ b/interface.md
@@ -210,8 +210,8 @@ therefore adapt the interface of
 `logdensityof` should suffice for variants, with the distinction being made by the capabilities of
 the concrete model instance.
 
- DensityInterface.jl also requires the trait function `hasdensity`, which is set to `true` for the
-`AbstractProbabilisticProgram` type.  Additional functions
+ DensityInterface.jl also requires the trait function `DensityKind`, which is set to `HasDensity()`
+for the `AbstractProbabilisticProgram` type.  Additional functions
 
 ```
 DensityInterface.densityof(d, x) = exp(logdensityof(d, x))
--- a/interface.md
+++ b/interface.md
@@ -29,7 +29,7 @@ Therefore, the interface consists of:
 - `condition(::Model, ::Trace) -> ConditionedModel`
 - `decondition(::ConditionedModel) -> GenerativeModel`
 - `sample(::Model, ::Sampler = Exact(), [Int])` (from `AbstractMCMC.sample`)
-- `logdensity(::Model, ::Trace)`
+- `DensityInterface.logdensityof(::Model, ::Trace)`
 
 
 ### Traces & probability expressions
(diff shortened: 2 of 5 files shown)


yebai:
> I think likelihood-free inference is out of scope for the kind of models we target -- is that a reasonable assumption?

That should be fine for now. We don't have any support for likelihood-free inference yet. But even for likelihood-free inference, one often uses an approximate likelihood/loss function. So we can still use the density interface.

yebai (approved):

interface.md
```diff
@@ -29,7 +29,7 @@ Therefore, the interface consists of:
 - `condition(::Model, ::Trace) -> ConditionedModel`
 - `decondition(::ConditionedModel) -> GenerativeModel`
 - `sample(::Model, ::Sampler = Exact(), [Int])` (from `AbstractMCMC.sample`)
-- `logdensity(::Model, ::Trace)`
+- `DensityInterface.logdensityof(::Model, ::Trace)`
```
devmotion, line 32: Maybe to be consistent with the line above:
```suggestion
- `logdensityof(::Model, ::Trace)` (from `DensityInterface.logdensityof`)
```
And maybe add and export `densityof` as well which is defined automatically?

interface.md
```diff
@@ -198,32 +198,34 @@ Not all variants need to be supported – for example, a posterior model might n
...
 Since the different “versions” of how a model is to be understood as generative or conditioned are
 to be expressed in the type or dispatch they support, there should be no need for separate functions
-`logjoint`, `loglikelihood`, etc., which force these semantic distinctions on the implementor; one
-`logdensity` should suffice for all, with the distinction being made by the capabilities of the
-concrete model instance.
+`logjoint`, `loglikelihood`, etc., which force these semantic distinctions on the implementor; we
+therefore adapt the interface of
+[DensityInterface.jl](https://github.com/JuliaMath/DensityInterface.jl).  Its one function
+`logdensityof` should suffice for variants, with the distinction being made by the capabilities of
+the concrete model instance.
```
devmotion, line 209: It's actually two functions (or three, if one includes `logfuncdensity` which seems irrelevant here though), maybe mention `densityof` as well?

src/AbstractPPL.jl
```diff
@@ -5,7 +5,7 @@ export VarName, getsym, getindexing, getlens, inspace, subsumes, varname, vsym,
 
 
 # Abstract model functions
-export AbstractProbabilisticProgram, condition, decondition, logdensity
+export AbstractProbabilisticProgram, condition, decondition, logdensityof
```
devmotion, line 8: Maybe
```suggestion
export AbstractProbabilisticProgram, condition, decondition, logdensityof, densityof
```

src/abstractprobprog.jl
```diff
@@ -8,19 +9,21 @@ Common base type for models expressed as probabilistic programs.
...
 
 `trace` can be of any supported internal trace type, or a fixed probability expression.
 
-`logdensity` should interact with conditioning and deconditioning in the way required by probability
-theory.
+`logdensityof` should interact with conditioning and deconditioning in the way required by
+probability theory.
 """
-function logdensity end
+DensityInterface.logdensityof
```
devmotion, line 26: I think it would be better to add a docstring only to `logdensityof(::AbstractProbabilisticProgram, ::AbstractModelTrace)`.
```suggestion
function DensityInterface.logdensityof(::AbstractProbabilisticProgram, ::AbstractModelTrace) end
```

devmotion (commented):
We should also update the version number :slightly_smiling_face:

src/abstractprobprog.jl
torfjelde: Should we not also deprecate `logdensity` rather than just dropping it completely?

devmotion: I used this form since `function f end` form is the recommended syntax in Julia documentation: https://docs.julialang.org/en/v1/manual/documentation/#Functions-and-Methods-2
This also adds a method to `f`.

It seems one can avoid this (even though it seems the other form is recommended) by specifying
```suggestion
DensityInterface.logdensityof(::AbstractProbabilisticProgram, ::AbstractModelTrace)
```
instead.

interface.md
```diff
@@ -198,32 +200,45 @@ Not all variants need to be supported – for example, a posterior model might n
...
 
-Note that this generalizes `logpdf`, too, since the posterior density will of course in general be
-unnormalized and hence not a probability density.
+ DensityInterface.jl also requires the trait function `hasdensity`, which is set to `true` for the
+`AbstractProbabilisticProgram` type.  Additional functions
 
-The evaluation will usually work with the internal, concrete trace type, like `VarInfo` in Turing.jl:
+```
+DensityInterface.densityof(d, x) = exp(logdensityof(d, x))
+DensityInterface.logdensity(d) = Base.Fix1(logdensity, d)
```
mschauer, line 218: Is this correct?

interface.md
```diff
@@ -24,12 +24,14 @@ There are three interrelating aspects that this interface intends to standardize
...
 
-Therefore, the interface consists of:
+Therefore, the interface consists of an `AbstractProbabilisticProgram` supertype, together with
+functions
 
 - `condition(::Model, ::Trace) -> ConditionedModel`
 - `decondition(::ConditionedModel) -> GenerativeModel`
 - `sample(::Model, ::Sampler = Exact(), [Int])` (from `AbstractMCMC.sample`)
-- `logdensity(::Model, ::Trace)`
+- `logdensityof(::Model, ::Trace)` and `density(::Model, ::Trace)` (from
```
mschauer, line 33: Is this correct?

src/AbstractPPL.jl
```diff
@@ -5,7 +5,7 @@ export VarName, getsym, getindexing, getlens, inspace, subsumes, varname, vsym,
 
 
 # Abstract model functions
-export AbstractProbabilisticProgram, condition, decondition, logdensity
+export AbstractProbabilisticProgram, condition, decondition, logdensityof, densityof
```
mschauer, line 8: Is densityof defined somewhere?

phipsgabler: Defaulted in DensityInterface, yes. https://github.com/JuliaMath/DensityInterface.jl/blob/ccade11dec6920cbc19852650e565d10254f1285/src/interface.jl#L156

interface.md
mschauer: No, it should be `densityof`

interface.md
mschauer: No it should be `logdensityof`

src/AbstractPPL.jl
mschauer: Ok

devmotion:
Yes, this is one of the main reasons for a unified interface. However, it will require some changes in DynamicPPL since we don't use `logpdf` currently but `loglikelihood` due to its implicit IID assumptions and support of multiple observations.

mschauer:
That's why I am asking, in the current form of https://github.com/JuliaStats/Distributions.jl/pull/1416 using `IIDDensity` in Turing will mean that the `logdensityof` interface is only supported for `B<:Distribution` because `IIDDensity` lives in Distributions.jl

interface.md
phipsgabler: Ah, right. I'm still getting used to the `of`...

mschauer: I was confused too

devmotion:
Yes, that's something we have to address in the future in some way 🙂 Currently, Distributions is baked in in DynamicPPL and Turing very deeply, so it will take some time and effort to generalize it regardless of `IIDDensity`.

phipsgabler:
TBH, I had no idea we used `loglikelihood`. Is that the use case? Is it used only for observations/conditioned variables?

```julia
julia> @model function foo(x)
           m ~ Normal()
           x ~ Normal(m)
       end
foo (generic function with 2 methods)

julia> foo(1.2)()
1.2

julia> foo([1.2, 0.2])()
2-element Vector{Float64}:
 1.2
 0.2
```

Is this really a good thing? It definitely makes the type checker in my brain complain.

devmotion:
It is a good idea since before we had to maintain all kinds of custom dispatches for different distributions (but of course would miss any others) and always allocated arrays for the individual log densities which then were summed immediately. It is much more sane to let it implement whoever defines a Distribution type.

devmotion:
But I would suggest not polluting this PR with the seemingly unrelated discussion 🙂

mschauer:
"Polluting" perhaps don't use that loaded language.

mschauer:
No problem, I just wanted to signal that the statement touched a nerve with me on the first take.

Project.toml
```diff
@@ -3,13 +3,15 @@ uuid = "7a57a42e-76ec-4ea3-a279-07e840d6d9cf"
 keywords = ["probablistic programming"]
 license = "MIT"
 desc = "Common interfaces for probabilistic programming"
-version = "0.3.1"
+version = "0.4"
```
devmotion, line 6: In principle the PR doesn't have to be breaking, you could just deprecate `logdensity`. But I guess DynamicPPL is the only/main dependency of AbstractPPL currently, so it doesn't really matter.

phipsgabler: As we are still in beta, and want to promote DensityInterface adoption, I prefer breaking.  We could leave `logdensity` deprecated in DPPL, though, I don't know how long it has been around before.

devmotion: Sounds good!

Issue #30, opened 2021-08-23
Setup `bors` workflow for AbstractPPL


yebai:
I suggest that we copy the `bors` workflow from `DynamicPPL`, which seems to work well.

PR #42, opened 2021-11-09
[Merged by Bors] - Setup `bors`

phipsgabler:
Implements #30. Is there anything else to set up for bors?

### DIFF
--- a/.github/workflows/Format.yml
+++ b/.github/workflows/Format.yml
@@ -9,6 +9,7 @@ on:
       - trying
       # Build the main branch.
       - main
+  pull_request:
 
 concurrency:
   # Skip intermediate builds: always.
--- a/.github/workflows/Format.yml
+++ b/.github/workflows/Format.yml
@@ -9,7 +9,6 @@ on:
       - trying
       # Build the main branch.
       - main
-  pull_request:
 
 concurrency:
   # Skip intermediate builds: always.
--- a/.github/workflows/Format.yml
+++ b/.github/workflows/Format.yml
@@ -1,0 +1,38 @@
+name: Format
+
+on:
+  push:
+    branches:
+      # This is where pull requests from "bors r+" are built.
+      - staging
+      # This is where pull requests from "bors try" are built.
+      - trying
+      # Build the main branch.
+      - main
+
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: only if it is a pull request build.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: 1
+      - name: Format code
+        run: |
+          using Pkg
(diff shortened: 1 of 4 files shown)


devmotion:
The Github CI has to be updated as well.

It might be a good idea also to make the bors test a mandatory check in the Github settings, so that you can accidentally merge a PR.

devmotion:
More concretely, one has to add https://github.com/TuringLang/DynamicPPL.jl/blob/ad545beaab3f104d62a01d926ad2d2998cb41614/.github/workflows/CI.yml#L5-L9 to all actions that bors should run. And one can probably remove the `pull_request:` from e.g. the CI action and integration tests such that these are only run when bors triggers them. It seems you also have to copy some of the actions (docs, format, etc.) if you include them in bors.toml. Otherwise bors will never merge any PRs :smile:

phipsgabler:
I don't really have a plan of what I'm doing here :D But I tried my best.

From the bors.toml, I guess the fourth argument (threads) should be removed, as we don't have that configuration here; and also the docs action. Right?

bors.toml
```diff
@@ -0,0 +1,20 @@
...
+  "test (1, ubuntu-latest, x86, 2)",
+  "test (1.%, ubuntu-latest, x64, 1)",
+  "test (1.%, ubuntu-latest, x64, 2)",
+  "test (1, ubuntu-latest, x64, 1)",
+  "test (1, ubuntu-latest, x64, 2)",
+  "test (1, macOS-latest, x64, 2)",
+  "test (1, windows-latest, x64, 2)",
+  "Turing.jl",
+  "format",
+  "docs"
```
devmotion, line 11: This should be a list of the names of the Github actions that are required to pass (`1.%` is a regex for the oldest supported Julia version, otherwise one would have to update it from eg. `1.0` to `1.1` manually):
```suggestion
  "test (1.%, ubuntu-latest, x64)",
  "test (1.%, macOS-latest, x64)",
  "test (1.%, windows-latest, x64)",
  "test (1, ubuntu-latest, x64)",
  "test (1, macOS-latest, x64)",
  "test (1, windows-latest, x64)",
  "DynamicPPL.jl",
  "format"
```

yebai:
@devmotion can you take a look at the `bors` setup, and see if there is anything missing before it gets merged? From now on, we would like to adopt the same workflow as DynamicPPL here.

.github/workflows/Format.yml
```diff
@@ -0,0 +1,38 @@
...
+on:
+  push:
+    branches:
+      # This is where pull requests from "bors r+" are built.
+      - staging
+      # This is where pull requests from "bors try" are built.
+      - trying
+      # Build the main branch.
+      - main
+
```
devmotion, line 12:
```suggestion
  pull_request:

```

devmotion (commented):
Looks good, we'll see if `bors try` or `bors r+` work.

devmotion:
bors try

devmotion:
bors try

.github/workflows/Format.yml
```diff
@@ -0,0 +1,39 @@
...
+on:
+  push:
+    branches:
+      # This is where pull requests from "bors r+" are built.
+      - staging
+      # This is where pull requests from "bors try" are built.
+      - trying
+      # Build the main branch.
+      - main
+  pull_request:
```
yebai, line 12:
```suggestion
```

yebai (approved):

Issue #35, opened 2021-09-20
More aggressive concretization


phipsgabler:
I propose to make `concretize` turn every `:` into the value of `axis(x, i)`.

I think this would make the set of `VarName`s we are working on a nicely defined poset, since then it's basically "sequences of tuples of sets" with the subsumption relation following inductively from sets of indices (leaving out linear indexing and trailing ones).

That should also make [this weird corner case](https://github.com/TuringLang/AbstractPPL.jl/blob/76f64dd9263988d4bea28806d3d8586be4d170bf/src/varname.jl#L419) obsolete.

And while we're at it: boolean indexing could be concretized to array indexing:

```julia
concretize(@varname(x[[true, false]], x) == @varname(x[[1,2]])
```

devmotion:
Would this break variables of dynamically changing size, similar to https://turing.ml/dev/tutorials/06-infinite-mixture-model/ where the size of `mu` changes in the for loop?

phipsgabler:
Not more than concretization of `begin` and `end` would, I assume?  All of these depend essentially only on the length/axes of the value at runtime.

But looking at this from the other side -- have we thoroughly tested the existing lens-based varname implementation for this case...?

phipsgabler:
I mean, even if you have the weirdest case I can think of, 

```
x[end] ~ D
push!(x, ...)
``` 

in a loop, it should be save, since the `end` is concretized at the point where the assignment happens and thus becomes `x[<current index in the loop>]`.  

And `:` you can only use legally once in a tilda, because otherwise, you'd be sampling the same location twice.  

Am I missing something?

phipsgabler:
@torfjelde what do you think about this?

PR #43, opened 2021-11-20
[Merged by Bors] - Concretize IndexLens using to_indices

phipsgabler:
Should implement #35.  The "problem" with this is that the resulting values are not exactly what you'd write by hand:

```julia
julia> AbstractPPL.concretize(@varname(x.a[1:end, end][:]), x)
x.a[1:2,2][Base.Slice(Base.OneTo(2))]
```

That wouldn't be so much of a problem besides printing, but unfortunately, lenses currenctly compare equality using strict types:

```julia
julia> AbstractPPL.concretize(@varname(x[1:3, :]), rand(10, 10)).lens.indices == @varname(x[1:3, 1:10]).lens.indices
true

julia> AbstractPPL.concretize(@varname(x[1:3, :]), rand(10, 10)).lens == @varname(x[1:3, 1:10]).lens
false

julia> AbstractPPL.concretize(@varname(x[1:3, :]), rand(10, 10)) == @varname(x[1:3, 1:10])
false
```

Cf. https://github.com/jw3126/Setfield.jl/issues/165; the equality comparison can hopefully be fixed there.

The remaining thing is that subsumption must still be able to work with `Colon`, since a user might index a trace/VarInfo using a non-concretized varname containing a colon.  But at least we can then be sure that one side is always concrete.


phipsgabler:
I added some [normalization code](https://github.com/TuringLang/AbstractPPL.jl/pull/43/files#diff-420ea19945b7708559a31e115d9c2139ec66f38492dc88ac84479ec83b30856c) to the index conversions, so that all ranges will become "normal ranges", as constructed by `:`.  

This should mitigate possible confusions in printing and problems with equality.  Also, at the time lenses are actually applied for getting, the indices are converted **once more** using `to_indices`, anyway, so the potential optimization of `OneTo` etc. is not lost.

torfjelde:
I'm generally okay with this change, but I'd like to think about it a tiny bit more. E.g. 

> The remaining thing is that subsumption must still be able to work with Colon, since a user might index a trace/VarInfo using a non-concretized varname containing a colon. But at least we can then be sure that one side is always concrete.

this is quite an annoying issue. Do you have any thoughts on how we'd approach this dealing with this?

src/varname.jl
```diff
@@ -418,28 +418,34 @@ _issubrange(i::ConcreteIndex, j::Colon) = true
 # we preserve the status quo, but I'm confused.
 _issubrange(i::Colon, j::ConcreteIndex) = true
 
+"""Turn any `AbstractRange` into an equivalent `UnitRange` or `StepRange`."""
+_normalizerange(r::AbstractRange) = (:)(r...)
```
devmotion, line 422: Is this normalization guaranteed to work and documented somewhere?

src/varname.jl
```diff
@@ -418,28 +418,34 @@ _issubrange(i::ConcreteIndex, j::Colon) = true
...
+
 """
     concretize(l::Lens, x)
 
 Return `l` instantiated on `x`, i.e. any runtime information evaluated using `x`.
 """
 concretize(I::Lens, x) = I
-concretize(I::DynamicIndexLens, x) = IndexLens(I.f(x))
+concretize(I::IndexLens, x) = IndexLens(_normalizerange.(Base.to_indices(x, I.indices)))
+concretize(I::DynamicIndexLens, x) = IndexLens(_normalizerange.(I.f(x)))
```
devmotion, line 432: No need for broadcasting here:
```suggestion
concretize(I::IndexLens, x) = IndexLens(map(_normalizerange, Base.to_indices(x, I.indices)))
concretize(I::DynamicIndexLens, x) = IndexLens(map(_normalizerange, I.f(x)))
```

devmotion (commented):
I guess this is quite breaking for DynamicPPL and Turing since the names of the parameters in the resulting chains etc. are based on this printed representation.

Probably even without this PR it would be good to test non-standard indices (e.g. OffsetArrays) - or is it already part of the test suite?

src/varname.jl
phipsgabler: Ha. It doesn't even work, my test cases were flawed :D

phipsgabler:
> > The remaining thing is that subsumption must still be able to work with Colon, since a user might index a trace/VarInfo using a non-concretized varname containing a colon. But at least we can then be sure that one side is always concrete.
> 
> this is quite an annoying issue. Do you have any thoughts on how we'd approach this dealing with this?

Sure, lets think about it. In my intuition, `:` in a varname without concrete context is a bit like infinity. If you have only of the, it works quite fine, you only run into undefined situations when there's two.

But my solution would be to do the same thing as the `to_index` family of functions, and always pass along the value you're talking about. `:` only means somthing _in the context_ of a given array. Then subsumption checks of `:` amount to subsumption of `to_index(x, :)`.

Now that's the logical answer. But if we do it this way, you compute the `to_index` at every lookup. Hence the idea to normalize it right away. Now, as @devmotion noted corretly, this changes printing, on which chains unfortunately depend... OTOH, we _already_ change printing: https://github.com/TuringLang/AbstractPPL.jl/blob/3897231e14242fe0dcc5c762ecea71fbac218dfc/src/varname.jl#L174. So one hacky solution would be to just continue with that.  For example, 

```julia
julia> to_indices([1,2], (:,))
(Base.Slice(Base.OneTo(2)),)
```

so we can dispatch on `Base.Slice`, which is used when a `Colon` is concretized.

devmotion:
> So one hacky solution would be to just continue with that.

There's a major difference here though: `:` is an alias of the singleton `Colon()` (https://github.com/JuliaLang/julia/blob/8c9799530383d571d5beab4e530d86e40a289929/base/essentials.jl#L691), and hence both are identical and can't be distinguished. Whereas this is clearly not the case for `Slice`.

phipsgabler:
I found that the following is also a problem, and more severe:

```julia
julia> y = rand(10, 10);

julia> to_indices(x, (2:2:5,))
(2:2:4,)
```

phipsgabler:
I've made some changes trying to find a balance here. Namely, 

- `to_indices` is applied to all indices, but values which were originally `Colon` are un-converted to just `UnitRange(converted_index)` (which is what you get when writing `begin:end` instead).
- I reintroduced the explicit `concretize` argument for `varname`, since now there is an actual choice to be made for "non-dynamic" varnames. This behaves the same as previously and fails on `begin`/`end`, just with a different error message.

I thought it would be possible to wrap the un-converted index into `Base.Slice`, but Base doesn't like `Slice{UnitRange}`. So the only reasonable thing IMHO is to do it as I described. This _does_ have the effect that stringification changes: `y[:]` becomes `y[1:N]`.

yebai:
@torfjelde can you take a look at this PR, see whether you are happy before I merge it?

phipsgabler:
I'd have to look at this one myself again; IIRC there were some unexpected subtleties.

phipsgabler:
Sooo... I think we need to think about this a bit more. I have now added a couple of proper tests for `VarName`, and fixed some edge cases. Setfield 0.8.1 is required to make things like

```julia
@varname(A[1, Not(3)], true) == @varname(A[1, [1, 2, 4, 5, 6, 7, 8, 9, 10]])
```

work.

I have started to continue and refactor some other things (so currently many subsumption tests are broken).

phipsgabler:
I got quite something, and I think it's reasonable. Please review.

src/varname.jl
```diff
@@ -297,10 +298,18 @@ subsumes(t::PropertyLens, u::PropertyLens) = false
...
+    u::Union{IndexLens,ComposedLens{<:IndexLens}}
+) = subsumes_index(t, u)
+
+
+const subsumedby = (t, u) -> subsumes(u, t)
+const ⊒ = subsumes
+const ⊑ = subsumedby
+const ⋣ = !subsumes
+const ⋢ = !subsumedby
+const ≍ = (t, u) -> t ⋢ u && u ⋢ t
```
devmotion, line 312: Are these Unicode operators defined by other packages as well which might cause conflicts? Are they needed?

devmotion, line 312:
```suggestion
≍(t, u) = t ⋢ u && u ⋢ t
```

src/varname.jl
```diff
@@ -396,80 +405,109 @@ Currently _not_ supported are:
...
-_issubrange(i::Colon, j::ConcreteIndex) = true
+Create the index to be emitted in `concretize`.  `original_index` is the original, unconcretized
+index, and `lowered_index` the respective position of the result of `to_indices`.
+
+The only purpose of this are special cases like `:`, which we want to avoid becoming a
+`Base.Slice(OneTo(...))` -- it would confuse people when printed.  Instead, we concretize to a
+`UnitRange` based on the `lowered_index`, just what you'd get with an explicit `begin:end`
+"""
+reconcretize_index(original_index, lowered_index) = lowered_index
+reconcretize_index(original_index::Colon, lowered_index::Base.Slice) = (UnitRange(lowered_index))
```
devmotion, line 433:
```suggestion
reconcretize_index(original_index::Colon, lowered_index::Base.Slice) = UnitRange(lowered_index)
```

src/varname.jl
```diff
@@ -502,38 +540,40 @@ julia> getlens(@varname(x.a))
...
 
-        return if Setfield.need_dynamic_lens(expr)
-            :(
+        if concretize
+            return :(
                 $(AbstractPPL.VarName){$(QuoteNode(sym))}(
                     $(AbstractPPL.concretize)($lens, $sym_escaped)
                 )
             )
+        elseif !concretize && Setfield.need_dynamic_lens(expr)
```
devmotion, line 570: `!concretize` is implied by the first branch?

```suggestion
        elseif Setfield.need_dynamic_lens(expr)
```

devmotion (commented):
Are the questions and possible issues in the discussion above resolved? If the PR is merged, I think it would be safer to put it into a breaking release due to the potential downstream breakages - or are they avoided in the latest version of the PR?

yebai:
> This assumes that non-concrete colons can never be subsumed (you cannot reasonably decide that x[:] is subsumed by x[1]; the other way around it makes sense), and so I error in this case.

Thanks, @phipsgabler for looking further into it.  I think it is sensible to throw an error on these corner cases. It is also fine to forbid cases where it causes inconsistency with downstream packages like `MCMCChains`, if that helps to keep the complexity of PR under control.

src/varname.jl
```diff
@@ -297,10 +298,18 @@ subsumes(t::PropertyLens, u::PropertyLens) = false
...
-subsumes(t::ComposedLens{<:IndexLens}, u::ComposedLens{<:IndexLens}) = subsumes_index(t, u)
-subsumes(t::IndexLens, u::ComposedLens{<:IndexLens}) = subsumes_index(t, u)
-subsumes(t::ComposedLens{<:IndexLens}, u::IndexLens) = subsumes_index(t, u)
+subsumes(
+    t::Union{IndexLens,ComposedLens{<:IndexLens}},
+    u::Union{IndexLens,ComposedLens{<:IndexLens}}
+) = subsumes_index(t, u)
+
+
+const subsumedby = (t, u) -> subsumes(u, t)
```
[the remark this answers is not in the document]
torfjelde: The same holds for the rest of the `const`, right? As in, we prefer `⊒(args...) = subsumes(args...)`, etc. too?

src/varname.jl
```diff
@@ -396,80 +405,109 @@ Currently _not_ supported are:
...
-function subsumes_index(t::Tuple, u::Tuple)  # does x[i]... subsume x[j]...?
-    return _issubindex(first(t), first(u)) && subsumes_index(Base.tail(t), Base.tail(u))
+function subsumes_index(t1::Tuple, t2::Tuple)  # does x[i]... subsume x[j]...?
+    first_subsumed = all(zip(first(t1), first(t2))) do (i, j)
+        if j isa Colon
+            error("Colons cannot be subsumed")
+        elseif i isa Colon
+            return true
+        else
+            return issubset(j, i)
```
torfjelde, line 415: This will fail for keys that aren't integers, e.g.

``` julia
julia> vn1 = @varname(x[:a]); vn2 = @varname(x[:a][1]);

julia> vn1 ≍ vn2
ERROR: MethodError: no method matching length(::Symbol)
Closest candidates are:
  length(::Union{Base.KeySet, Base.ValueIterator}) at /opt/julia-1.7.0/share/julia/base/abstractdict.jl:58
  length(::Union{LinearAlgebra.Adjoint{T, S}, LinearAlgebra.Transpose{T, S}} where {T, S}) at /opt/julia-1.7.0/share/julia/stdlib/v1.7/LinearAlgebra/src/adjtrans.jl:171
  length(::Union{Tables.AbstractColumns, Tables.AbstractRow}) at ~/.julia/packages/Tables/M26tI/src/Tables.jl:175
  ...
Stacktrace:
  [1] issubset(l::Symbol, r::Symbol)
    @ Base /opt/julia-1.7.0/share/julia/base/abstractset.jl:281
  [2] (::AbstractPPL.var"#15#16")(::Tuple{Symbol, Symbol})
    @ AbstractPPL ~/Projects/public/AbstractPPL.jl/src/varname.jl:415
  [3] _all(f::AbstractPPL.var"#15#16", itr::Base.Iterators.Zip{Tuple{Tuple{Symbol}, Tuple{Symbol}}}, #unused#::Colon)
    @ Base /opt/julia-1.7.0/share/julia/base/reduce.jl:1157
  [4] all(f::Function, itr::Base.Iterators.Zip{Tuple{Tuple{Symbol}, Tuple{Symbol}}})
    @ Base /opt/julia-1.7.0/share/julia/base/reduce.jl:1152
  [5] subsumes_index(t1::Tuple{Tuple{Symbol}, Tuple{Int64}}, t2::Tuple{Tuple{Symbol}})
    @ AbstractPPL ~/Projects/public/AbstractPPL.jl/src/varname.jl:409
  [6] subsumes_index(t::Setfield.ComposedLens{Setfield.IndexLens{Tuple{Symbol}}, Setfield.IndexLens{Tuple{Int64}}}, u::Setfield.IndexLens{Tuple{Symbol}})
    @ AbstractPPL ~/Projects/public/AbstractPPL.jl/src/varname.jl:361
  [7] subsumes(t::Setfield.ComposedLens{Setfield.IndexLens{Tuple{Symbol}}, Setfield.IndexLens{Tuple{Int64}}}, u::Setfield.IndexLens{Tuple{Symbol}})
    @ AbstractPPL ~/Projects/public/AbstractPPL.jl/src/varname.jl:304
  [8] subsumes(u::VarName{:x, Setfield.ComposedLens{Setfield.IndexLens{Tuple{Symbol}}, Setfield.IndexLens{Tuple{Int64}}}}, v::VarName{:x, Setfield.IndexLens{Tuple{Symbol}}})
    @ AbstractPPL ~/Projects/public/AbstractPPL.jl/src/varname.jl:269
  [9] subsumedby(t::VarName{:x, Setfield.IndexLens{Tuple{Symbol}}}, u::VarName{:x, Setfield.ComposedLens{Setfield.IndexLens{Tuple{Symbol}}, Setfield.IndexLens{Tuple{Int64}}}})
    @ AbstractPPL ~/Projects/public/AbstractPPL.jl/src/varname.jl:307
 [10] ⋢(t::VarName{:x, Setfield.IndexLens{Tuple{Symbol}}}, u::VarName{:x, Setfield.ComposedLens{Setfield.IndexLens{Tuple{Symbol}}, Setfield.IndexLens{Tuple{Int64}}}})
    @ AbstractPPL ~/Projects/public/AbstractPPL.jl/src/varname.jl:311
 [11] ≍(t::VarName{:x, Setfield.IndexLens{Tuple{Symbol}}}, u::VarName{:x, Setfield.ComposedLens{Setfield.IndexLens{Tuple{Symbol}}, Setfield.IndexLens{Tuple{Int64}}}})
    @ AbstractPPL ~/Projects/public/AbstractPPL.jl/src/varname.jl:312
 [12] top-level scope
    @ REPL[14]:1
```

More specifically

``` julia
julia> issubset(:a, :a)
ERROR: MethodError: no method matching length(::Symbol)
Closest candidates are:
  length(::Union{Base.KeySet, Base.ValueIterator}) at /opt/julia-1.7.0/share/julia/base/abstractdict.jl:58
  length(::Union{LinearAlgebra.Adjoint{T, S}, LinearAlgebra.Transpose{T, S}} where {T, S}) at /opt/julia-1.7.0/share/julia/stdlib/v1.7/LinearAlgebra/src/adjtrans.jl:171
  length(::Union{Tables.AbstractColumns, Tables.AbstractRow}) at ~/.julia/packages/Tables/M26tI/src/Tables.jl:175
  ...
Stacktrace:
 [1] issubset(l::Symbol, r::Symbol)
   @ Base /opt/julia-1.7.0/share/julia/base/abstractset.jl:281
 [2] top-level scope
   @ REPL[15]:1
```

Though this should also be broken in the implementation currently present on `#master`.

src/varname.jl
```diff
@@ -396,80 +405,109 @@ Currently _not_ supported are:
...
 
-const AnyIndex = Union{Int,AbstractVector{Int},Colon}
-_issubindex_(::Tuple{Vararg{AnyIndex}}, ::Tuple{Vararg{AnyIndex}}) = false
-function _issubindex(t::NTuple{N,AnyIndex}, u::NTuple{N,AnyIndex}) where {N}
-    return all(_issubrange(j, i) for (i, j) in zip(t, u))
-end
 
-const ConcreteIndex = Union{Int,AbstractVector{Int}} # this include all kinds of ranges
+"""
+    concretize_index(original_index, lowered_index)
```
torfjelde, line 423:
```suggestion
    reconcretize_index(original_index, lowered_index)
```

src/varname.jl
torfjelde: One solution is to split the if-statement into different dispatches for some function, and then add a default `f(x, y) = x == y` or something.

torfjelde (changes requested):

src/varname.jl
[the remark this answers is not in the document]
devmotion: No, in the other cases (apart from the last one) we want the Unicode character to be an alias such that we can use the ASCII and Unicode names interchangeably. Otherwise we define a new function that just forwards all arguments to another function but dispatches, ChainRules definitions etc. of one of these functions wouldn't cover the other one.

src/varname.jl
phipsgabler: I know, this is somethink I monkey-patched again to make tests work. Originally I wanted `checkindex`, but that unfortunately throws an error in cases it is not meant for and is even worse. Dispatch sounds good; maybe we can come up with a reasonably extensible interface here.

src/varname.jl
phipsgabler: I certainly think in terms of `⊑`, because I don't intuit the argument order of `subsumes`.  And then the rest follows logically.

The question about other packages is a good one; any reasonable way to find out about that besides asking on Slack?

phipsgabler: BTW I'm not sure about the negated operators. They seem nice to have, but one issue is that they are appearently rendered wrong in some fonts; I see them as `\nsubset`, not `\nsubseteq`, i.e., without the line underneath.

phipsgabler:
What remains to be discussed are the choices about colons.  Is concretizing them into ranges OK, and do the subsumption rules involving them not cause troubles.

I think I can argue for both (and think that _some_ breaking changes are acceptable if it makes the overall implementation better and more consistent), but this needs to be cleared by people who can judge the downstream consequences better than me.

I'll add more test cases and README documentation to lay out the behaviour later this week.

src/varname.jl
devmotion: Isn't the order the same as `isa`, `issubset`, `in` etc.? Arguably if you read it as infix operator the order of the arguments is quite natural.

Generally, I think Unicode aliases are fine if there are no conflicts with major packages (maybe could be checked with juliahub.com) and if there are ASCII alternatives. Unicode is less accessible (it requires special support for displaying and in particular editing) and therefore in my opinion one should follow the practice in Julia base and always provide an ASCII alternative if some Unicode name is part of the API.

phipsgabler: Sure, that's why I kept `subsumes` and `subsumedby`.  I'll see whether I can find out about competing definitions.

phipsgabler:
I have a compromise suggestion: instead of not concretizing `:` or completely concretizing it to a `UnitRange`, we introduce a new type which clones `Base.Slice` (which is what `to_indices` of `:` results in), but prints as `:`:

```julia
julia> y = rand(10, 10);

julia> x = (;y);

julia> @varname(x.y[1, :])
x.y[1,:]

julia> @varname(x.y[1, :], true)
x.y[1,:]

julia> AbstractPPL.getlens(@varname(x.y[1, :], true)).inner.indices
(1, :)

julia> AbstractPPL.getlens(@varname(x.y[1, :], true)).inner.indices[2]
ConcretizedSlice(Base.OneTo(10))
```

That way, the colon is concretized, but not visibly distinguishable from `:`.  `ConcretizedSlice` should behave exactly like the original colon in indexing, and at the same time like the resulting `Base.Slice`.

src/varname.jl
```diff
@@ -383,93 +393,144 @@ function combine_indices(lens::ComposedLens{<:IndexLens})
...
+
+
+struct ConcretizedSlice{T, R} <: AbstractVector{T}
+    range::R
 end
 
-const ConcreteIndex = Union{Int,AbstractVector{Int}} # this include all kinds of ranges
+ConcretizedSlice(s::Base.Slice{R}) where {R} = ConcretizedSlice{eltype(s.indices), R}(s.indices)
+Base.show(io::IO, s::ConcretizedSlice) = print(io, ":")
+Base.show(io::IO, ::MIME"text/plain", s::ConcretizedSlice) = print(io, "ConcretizedSlice($(s.range))")
```
devmotion, line 428:
```suggestion
Base.show(io::IO, ::MIME"text/plain", s::ConcretizedSlice) = print(io, "ConcretizedSlice(", s.range, ")")
```

src/varname.jl
```diff
@@ -383,93 +393,144 @@ function combine_indices(lens::ComposedLens{<:IndexLens})
...
+Base.IteratorEltype(::Type{<:ConcretizedSlice}) = Base.HasEltype()
+Base.eltype(s::ConcretizedSlice{T}) where {T} = T
+Base.IteratorSize(::Type{<:ConcretizedSlice}) = Base.HasLength()
+Base.length(s::ConcretizedSlice) = length(s.range)
+Base.size(s::ConcretizedSlice) = size(s.range)
+Base.iterate(s::ConcretizedSlice, state...) = Base.iterate(s.range, state...)
+Base.collect(s::ConcretizedSlice) = collect(s.range)
+Base.hasfastin(::Type{<:ConcretizedSlice}) = true
+Base.in(i, s::ConcretizedSlice) = i in s.range
```
devmotion, line 437: Most of these definitions should be provided by the defaults for `AbstractVector` automatically. E.g., I don't think it's necessary to define `eltype`, `IteratorEltype`, and `IteratorSize`, and `length` is defined automatically by `size`.

src/varname.jl
devmotion: > They seem nice to have, but one issue is that they are appearently rendered wrong in some fonts; I see them as \\nsubset, not \\nsubseteq, i.e., without the line underneath.

I just noticed that they are displayed correctly in the Github app on my phone but incorrectly without the line underneath in Firefox on my linux laptop.

phipsgabler: Core.Compiler and JET use them for type lattice stuff (pretty much same motivation I applied), and BasicBSpline.jl for (I think) interval containment: https://juliahub.com/ui/Search?q=%E2%8A%91&type=code&f=.%2a%5C.jl

yebai:
bors try

yebai:
Bors try

yebai:
Some tests fail on Julia 1.0, which is probably OK to ignore. I updated the CI scripts to replace `1.0` with `1.6`, the new long term support Julia version.

devmotion:
It seems the PR breaks https://github.com/TuringLang/DynamicPPL.jl/blob/aeb5e03fdfe6958e4f79d9ccc5b01b84b93371ff/test/compiler.jl#L220. The model seems reasonable though, doesn't it?

Project.toml
```diff
@@ -14,5 +14,5 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 [compat]
 AbstractMCMC = "2, 3"
 DensityInterface = "0.4"
-Setfield = "0.7.1, 0.8"
-julia = "1"
+Setfield = "0.8.1"
+julia = "1.6, 1.7"
```
devmotion, line 18: This is equivalent to
```suggestion
julia = "1.6"
```

phipsgabler:
> It seems the PR breaks [TuringLang/DynamicPPL.jl@`aeb5e03`/test/compiler.jl#L220](https://github.com/TuringLang/DynamicPPL.jl/blob/aeb5e03fdfe6958e4f79d9ccc5b01b84b93371ff/test/compiler.jl#L220). The model seems reasonable though, doesn't it?

It does indeed. I need to take a look at that.

phipsgabler:
bors try

phipsgabler:
bors try

Project.toml
```diff
@@ -13,7 +13,7 @@ Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
-AbstractMCMC = "2, 3, 4"
+AbstractMCMC = "2, 3"
```
devmotion, line 16: I don't think we want to drop support for AbstractMCMC 4? Maybe rather drop support for < 4.
```suggestion
AbstractMCMC = "2, 3, 4"
```

Project.toml
```diff
@@ -13,7 +13,7 @@ Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
...
 DensityInterface = "0.4"
-Setfield = "0.7.1, 0.8"
-julia = "~1.6.6, 1.7.3"
+Setfield = "0.8.1"
+julia = "1.6"
```
devmotion, line 19: This reenables support for Julia 1.6.0-1.6.5 it seems? Is this desired?
```suggestion
julia = "~1.6.6, 1.7.3"
```

bors.toml
```diff
@@ -2,9 +2,9 @@ status = [
   "Julia 1.% - ubuntu-latest - x64",
   "Julia 1.% - macOS-latest - x64",
   "Julia 1.% - windows-latest - x64",
-  "Julia 1 - ubuntu-latest - x64",
-  "Julia 1 - macOS-latest - x64",
-  "Julia 1 - windows-latest - x64",
+  "Julia 1.6 - ubuntu-latest - x64",
+  "Julia 1.6 - macOS-latest - x64",
+  "Julia 1.6 - windows-latest - x64",
```
devmotion, line 7: This change seems wrong, that case is already covered above. Changing this means that it will pass regardless of whether tests fail with the latest stable version of Julia or not.
```suggestion
  "Julia 1 - ubuntu-latest - x64",
  "Julia 1 - macOS-latest - x64",
  "Julia 1 - windows-latest - x64",
```

phipsgabler:
The remaining DPPL errors are, I think, simply caused by the fact that `varname` should now be called with `concretize = true` in the compiler context.

Project.toml
phipsgabler: Ah dammit, there was something weird going on during mergeing main.

yebai (approved):

src/varname.jl
```diff
@@ -503,38 +567,40 @@ julia> getlens(@varname(x.a))
...
     if Meta.isexpr(expr, :ref) || Meta.isexpr(expr, :.)
         # Split into object/base symbol and lens.
         sym_escaped, lens = Setfield.parse_obj_lens(expr)
         # Setfield.jl escapes the return symbol, so we need to unescape
         # to call `QuoteNode` on it.
         sym = drop_escape(sym_escaped)
 
-        return if Setfield.need_dynamic_lens(expr)
-            :(
+        if concretize
```
torfjelde, line 591: @phipsgabler Do you remember why this change was made?

torfjelde: In particular, why it was made `false` by default.

PR #44, opened 2021-11-29
DAG representation using SimplePPL/Mamba

PavanChaggar:
The aims of this PR are to build a minimal interface for DAG representations of models. In the first instance, we'll try to do this using code from SimplePPL/Mamba. 

A SimplePPL `Model` currently has the following structure: 

```
mutable struct Model
    nodes::Dict{Symbol,Any}
    samplers::Vector{Sampler}
    VariableSet::Vector{VariableSet}
    states::Vector{ModelState}
    iter::Int
    burnin::Int
    hasinputs::Bool
    hasinits::Bool
end
```

This is very similar to the Mamba `Model`, with the exception of `VariableSet` which is used for defining graph partitions. Mamba models are tightly coupled with samplers. The first step is to completely decouple the DAG api from the samplers. The proposed `Model` struct is:

```
mutable struct Model
    nodes::Dict{Symbol,Any}
    hasinputs::Bool
    hasinits::Bool
end
```

We can also discuss removing `hasinputs` and `hasinits` in favour of using `ConditionedModel`. Also, would we like a separate field for the DAG? At present, we need to call `ModelGraph` on the `Model` to generate the graph. `ModelGraph` is called in the `Model` constructor to get dependent and target node information but the DAG itself is not saved.

After the Model and samplers have been decoupled, we'll simplify the `Node` data structures. At the moment, there is a lot of code repetition and specialised dispatch on node types. @yebai suggested this could be simplified using NamedTuples.


Project.toml
```diff
@@ -3,11 +3,13 @@ uuid = "7a57a42e-76ec-4ea3-a279-07e840d6d9cf"
 keywords = ["probablistic programming"]
 license = "MIT"
 desc = "Common interfaces for probabilistic programming"
-version = "0.4"
+version = "0.4.0"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 DensityInterface = "b429d917-457f-4dbc-8f4c-0cc954292b1d"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
```
yebai, line 11: We can depend on the [`DensityInterface`](https://github.com/JuliaMath/DensityInterface.jl) instead of Distributions.

Project.toml
```diff
@@ -3,11 +3,13 @@ uuid = "7a57a42e-76ec-4ea3-a279-07e840d6d9cf"
...
+Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
```
yebai, line 12: Not sure how much functionality we need from `Graphs` - in general, we want to keep `AbstractPPL` very lightweight and remove dependency on `Graphs`.

src/simpleppl.jl
```diff
@@ -0,0 +1,749 @@
...
+    ::Union{Type{Array{T}},Type{Array{T,N}}},
+    v::ArrayVariate{N},
+) where {T<:Real,N} = convert(Array{T,N}, v.value)
+
+Base.unsafe_convert(::Type{Ptr{Float64}}, v::ArrayVariate) = pointer(v.value)
+
+
+macro promote_scalarvariate(V)
+    quote
+        Base.promote_rule(::Type{$(esc(V))}, ::Type{T}) where {T<:Real} = Float64
```
devmotion, line 94: This seems wrong in general - I don't think dual numbers or eg `BigFloat` should be promoted to `Float64`.

src/simpleppl.jl
```diff
@@ -0,0 +1,749 @@
...
+
+Base.stride(v::ArrayVariate, k::Int) = stride(v.value, k)
+
+
+#################### Indexing ####################
+
+Base.getindex(v::ScalarVariate, ind::Int) = v.value[ind]
+
+Base.getindex(v::ScalarVariate, inds::Union{StepRange{Int,Int},Vector{Int}}) =
+    Float64[v[i] for i in inds]
```
devmotion, line 111: This should not be needed.
```suggestion
```

src/simpleppl.jl
```diff
@@ -0,0 +1,749 @@
...
+
+Base.getindex(v::ArrayVariate, inds::Int...) = getindex(v.value, inds...)
```
devmotion, line 113: I guess if you define the `IndexStyle` as `IndexLinear` (might not be possible) then this should not be needed either.

src/simpleppl.jl
```diff
@@ -0,0 +1,749 @@
...
+) where {T<:Real}
+    nx = length(x)
+    ninds = length(inds)
+    nx == ninds ||
+        throw(DimensionMismatch("tried to assign $nx elements to $ninds destinations"))
+
+    for i = 1:nx
+        v[inds[i]] = x[i]
+    end
+end
```
devmotion, line 131: This should not be needed.

```suggestion
```

src/simpleppl.jl
```diff
@@ -0,0 +1,749 @@
...
+
+Base.setindex!(v::ArrayVariate, x, inds::Int...) = setindex!(v.value, x, inds...)
+
+
+#################### I/O ####################
+
+function Base.show(io::IO, v::AbstractVariate)
```
devmotion, line 138: The standard `show` method should return a compact single-line representation. You can define
```julia
show(::IO, ::MIME"text/plain", ::AbstractVariate)
```
for a more verbose multi-line representation.

src/simpleppl.jl
```diff
@@ -0,0 +1,749 @@
...
+    values = similar(v.value, AbstractString)
+    for i = 1:length(v)
+        s = string(ind2sub(size(v), i))
+        values[i] = string(prefix, "[", s[2:(end-offset)], "]")
+    end
+    values
+end
+
+
+#################### Mathematical Operators ####################
```
devmotion, line 160: This section highlights the problem with defining a custom `Real` number type: one has to define all supported operations that don't work automatically. This is annoying and makes it less likely that it just works with other functions (which could also be defined in other packages).

src/simpleppl.jl
```diff
@@ -0,0 +1,749 @@
...
+
+for op in UnaryScalarMethods
+    @eval ($op)(x::ScalarVariate) = ($op)(x.value)
+end
+
+  
+#################### Model Types ####################
+
+struct ModelGraph
+    graph::DiGraph
```
devmotion, line 216: Is this also an abstract type?

src/simpleppl.jl
```diff
@@ -0,0 +1,749 @@
...
+end
+
+#################### Dependent ####################
+
+const depfxargs = [(:model, Model)]
+
+
+#################### Base Methods ####################
+
+function Base.show(io::IO, d::AbstractDependent)
```
devmotion, line 233: This should be a compact representation.

phipsgabler:
So this is approximately what we're talking about:

```julia
m.value == (a = [0, 1], b = 0, c = 0)
m.input == (a = (), b = (), c = (:a, :b))
m.eval == (a = () -> D(), b = () -> 42, c = (a, b) -> Dist(a, 2sqrt(b)))
m.kind == (a = Stochastic, b = Logical, c = Stochastic)
m[@varname(a)] == (value = [0, 1], input = (), eval = ..., kind = Stochastic)
# @varname(a)::VarName{:a, IdentityLens}

# and perhaps
vn = @varname(a[1])
getvalue(m, vn) == get(m.value[getsym(vn)], getlens(vn)) == 0
```

devmotion (commented):
I added some comments.

In general, I wonder if AbstractPPL is the right place for this implementation. The package is supposed to be a lightweight interface package, as far as I understand, but the PR adds quite a lot of code and also not really lightweight dependencies such as Distributions and Graphs.

yebai:
> In general, I wonder if AbstractPPL is the right place for this implementation. The package is supposed to be a lightweight interface package, as far as I understand, but the PR adds quite a lot of code and also not really lightweight dependencies such as Distributions and Graphs.

We plan to keep `AbstractPPL` lightweight (see my comments above).  This PR is a WIP -- most code will be removed.

PavanChaggar:
closing this because the diff has become a bit too messy -- my bad!

PR #47, opened 2022-01-10
[Merged by Bors] - DAG Model interface

PavanChaggar:
This is a draft PR introducing a `Model` type that stores and makes use the model graph. 

The main type introduced here is the `Model` struct which stores the `ModelState` and `DAG`, each of which are their own types. `ModelState` contains information about the node values, dependencies and eval functions and `DAG` contains the graph and topologically ordered vertex list. 

A model can be constructed in the following way: 

```julia
julia> nt = (
               s2 = (0.0, (), () -> InverseGamma(2.0,3.0), :Stochastic), 
               μ = (1.0, (), () -> 1.0, :Logical), 
               y = (0.0, (:μ, :s2), (μ, s2) -> MvNormal(μ, sqrt(s2)), :Stochastic)
           )
(s2 = (0.0, (), var"#33#36"(), :Stochastic), μ = (1.0, (), var"#34#37"(), :Logical), y = (0.0, (:μ, :s2), var"#35#38"(), :Stochastic))

julia> Model(nt)
Nodes: 
μ = (value = 1.0, input = (), eval = var"#16#19"(), kind = :Logical)
s2 = (value = 0.0, input = (), eval = var"#15#18"(), kind = :Stochastic)
y = (value = 0.0, input = (:μ, :s2), eval = var"#17#20"(), kind = :Stochastic)
DAG: 
3×3 SparseMatrixCSC{Float64, Int64} with 2 stored entries:
  ⋅    ⋅    ⋅ 
  ⋅    ⋅    ⋅ 
 1.0  1.0   ⋅ 
```

At present, only functions needed for the constructors are implemented, as well as indexing using `@varname`. I still need to complete the integration with the AbstractPPL api. TODO: 
~~- [ ] `condition`/`decondition`,~~
~~- [ ] `sample`~~
~~- [ ] `logdensityof`~~
- [x] pure functions for ordered dictionary, as outlined in [AbstractPPL](https://github.com/TuringLang/AbstractPPL.jl#property-interface)

Feedback on `Model` structure welcome whilst I implement the remaining features!


cpfiffer:
No substantive comments other than that this is cool! May do a review later when I get more time.

src/simpleppl.jl
```diff
@@ -0,0 +1,213 @@
...
+julia> AbstractPPL.adjacency_matrix(inputs)
+3×3 SparseMatrixCSC{Float64, Int64} with 2 stored entries:
+  ⋅    ⋅    ⋅ 
+  ⋅    ⋅    ⋅ 
+ 1.0  1.0   ⋅ 
+"""
+function adjacency_matrix(inputs::NamedTuple)
+    N = length(inputs)
+    nodes = keys(inputs)
+    A = spzeros(N,N)
```
phipsgabler, line 112:
```suggestion
    A = spzeros(Bool, N,N)
```

src/simpleppl.jl
```diff
@@ -0,0 +1,213 @@
...
+    for (i, node) in enumerate(nodes)
+        v_inputs = inputs[node] # try as vector 
+       if v_inputs != () 
+            for inp in v_inputs
+                ind = findall(x -> x == inp, nodes)
+                A[i, ind[1]] = 1
```
phipsgabler, line 118:
```suggestion
                A[i, ind] = true
```

(A consequence of the above suggestions.)

src/simpleppl.jl
```diff
@@ -0,0 +1,213 @@
...
+  ⋅    ⋅    ⋅ 
+ 1.0  1.0   ⋅ 
+"""
+function adjacency_matrix(inputs::NamedTuple)
+    N = length(inputs)
+    nodes = keys(inputs)
+    A = spzeros(N,N)
+    for (i, node) in enumerate(nodes)
+        v_inputs = inputs[node] # try as vector 
+       if v_inputs != ()
```
phipsgabler, line 115: You can remove the `if` -- when there are no parents, the loop below is empty anyway.

src/simpleppl.jl
```diff
@@ -0,0 +1,213 @@
...
+"""
+function adjacency_matrix(inputs::NamedTuple)
+    N = length(inputs)
+    nodes = keys(inputs)
+    A = spzeros(N,N)
+    for (i, node) in enumerate(nodes)
+        v_inputs = inputs[node] # try as vector 
+       if v_inputs != () 
+            for inp in v_inputs
+                ind = findall(x -> x == inp, nodes)
```
phipsgabler, line 117:
```suggestion
                ind = findfirst(==(inp), nodes)
```

`findfirst` gives you the first index, or `nothing` in case there is none, which should raise an "unexpected" error.

So somewhere before, there should be a check of the `inputs` for consistency (i.e., all parents exist as keys).

src/simpleppl.jl
```diff
@@ -0,0 +1,213 @@
...
+    end
+    A
+end
+
+adjacency_matrix(m::Model) = adjacency_matrix(m.ModelState.input)
+
+function outneighbors(A::SparseMatrixCSC, u::T) where T <: Int
+    #adapted from Graph.jl
+    a = Array(A[:,u])
+    findall(x->x==1, a)
```
phipsgabler, line 130:
```suggestion
    inds, _ = findnz(A[:, u])
    inds
```

https://docs.julialang.org/en/v1/stdlib/SparseArrays/#SparseArrays.findnz.

src/simpleppl.jl
```diff
@@ -0,0 +1,213 @@
...
+end
+
+function topological_sort_by_dfs(A)
```
phipsgabler, line 133: Do you know which algorithm that is? A link to a reference would be good here.

src/simpleppl.jl
```diff
@@ -0,0 +1,213 @@
...
+julia> inputs = (a = (), b = (), c = (:a, :b))
+(a = (), b = (), c = (:a, :b))
+
+julia> AbstractPPL.adjacency_matrix(inputs)
+3×3 SparseMatrixCSC{Float64, Int64} with 2 stored entries:
+  ⋅    ⋅    ⋅ 
+  ⋅    ⋅    ⋅ 
+ 1.0  1.0   ⋅ 
+"""
+function adjacency_matrix(inputs::NamedTuple)
```
phipsgabler, line 109: I think you can remove the type constraint. Then it automatically works for, e.g., dictionaries, too.

src/simpleppl.jl
```diff
@@ -0,0 +1,213 @@
...
+"""
+struct DAG
+    A::SparseMatrixCSC
+    sorted_vertex_list::Tuple
+    
+    function DAG(in::NamedTuple) 
+        inputs = keys(in)
+        A = adjacency_matrix(in) 
+        v_list = topological_sort_by_dfs(A)
+        new(A, inputs[v_list])
```
phipsgabler, line 40:
```suggestion
    function DAG(inputs::NamedTuple) 
        input_names = keys(inputs)
        A = adjacency_matrix(inputs) 
        sorted_vertices = topological_sort_by_dfs(A)
        new(A, inputs[sorted_vertices])
```

Or something like that. At least, some better names.

Also, this should be an outer constructor. You can probably remove the type constraint, then it works also on dictionaries.

Then there should be some check of the inputs for consistency (existence of parent nodes etc.)

src/simpleppl.jl
```diff
@@ -0,0 +1,213 @@
...
+
+"""
+    DAG(inputs::NamedTuple{T})
+
+Struct containing the adjacency matrix for a particular model and 
+the topologically ordered vertex list.
+"""
+struct DAG
+    A::SparseMatrixCSC
+    sorted_vertex_list::Tuple
```
phipsgabler, line 34:
```suggestion
    sorted_vertices::Vector{Symbol}
```

src/simpleppl.jl
phipsgabler: And does the order of rows in `A` really correspond to the sorted order of vertices here? I think it better should, otherwise this looks a bit inconsistent.

Maybe you need to use [`SparseArrays.permute`](https://docs.julialang.org/en/v1/stdlib/SparseArrays/#SparseArrays.permute) after sorting.

phipsgabler (changes requested):

src/simpleppl.jl
PavanChaggar: http://www.cs.nott.ac.uk/~psznza/G5BADS03/graphs2.pdf

this is what's given in Graphs.jl [here](https://github.com/JuliaGraphs/Graphs.jl/blob/06669054ed470bcfe4b2ad90ed974f2e65c84bb6/src/traversals/dfs.jl)

src/simpleppl.jl
PavanChaggar: Thanks. The `A` matrix before did not correspond to the ordered vertex list -- nice catch. Fixed this now. Will work on adding some checks for correct graph structures.

src/simpleppl.jl
phipsgabler: Ah. Then just add a comment with link to the function in Graphs.jl.

PavanChaggar:
I think all the necessary functions needed for dictionary-like iterations are now supported. 

There was a problem with the adjacency matrix function in that it didn't support nodes that had only one input. I've added a fix for this but it's not totally type stable so still working on it.

yebai:
Just FYI: Julia now provides a standard way of obtaining argument names of methods: 

```julia
julia> f = (x, y) -> println(x, y)
#3 (generic function with 1 method)

julia> args = Base.method_argnames(first(methods(f)))
3-element Vector{Symbol}:
 Symbol("#self#")
 :x
 :y

julia> args[2:end]
2-element Vector{Symbol}:
 :x
 :y
```

Maybe we can make use of this, and provide an alternative `Model` constructor with less verbose arguments: 

```julia
julia> nt = (
               s2 = (0.0, () -> InverseGamma(2.0,3.0), :Stochastic), 
               μ = (1.0, () -> 1.0, :Logical), 
               y = (0.0, (μ, s2) -> MvNormal(μ, sqrt(s2)), :Stochastic) # (:μ, :s2) can be extracted using code above. 
           )


```

We can keep the verbose version as the internal API, and export the simpler version to users.

PavanChaggar:
Thanks for pointing that out @yebai, that would be much nicer. Is that just in Julia 1.7?

yebai:
> Thanks for pointing that out @yebai, that would be much nicer. Is that just in Julia 1.7?

It works for Julia 1.6 too. But I haven't checked lower Julia versions.

src/simpleppl.jl
```diff
@@ -113,26 +113,26 @@ julia> AbstractPPL.adjacency_matrix(inputs)
...
-    A
+    return A
 end
 
-function setinput!(A::SparseMatrixCSC{Bool, Int64}, row_n, nodes, v_inputs::Symbol)
-    ind = findfirst(==(v_inputs), nodes)
-    A[row_n, ind] = true
+function setinput!(A::SparseMatrixCSC{Bool, Int64}, row, col_inds, v_input::Symbol)
+    col = col_inds[v_input]
+    A[row, col] = true
```
phipsgabler, line 132: If you are not using this function anywhere else, I'd just inline it into `adjacency_matrix`.

src/simpleppl.jl
```diff
@@ -0,0 +1,231 @@
...
+`input` stores a tuple of inputs for a given node. `eval` are the
+anonymous functions associated with each node. These might typically
+be either deterministic values or some distribution, but could an 
+arbitrary julia program. `kind` is a tuple of symbols indicating
+whether the node is a logical or stochastic node.
+"""
+struct ModelState{T}
+    value::NamedTuple{T}
+    input::NamedTuple{T}
+    eval::NamedTuple{T}
```
yebai, line 22: Maybe we should consider moving the `eval` and `kind` fields into `DAG`? These fields actually provide the edge information (`eval` arguments implicitly codes the edges between nodes) and the node attribute (`logical` or `stochastic`) of the model graph. In addition, these fields remain constant during inference. They are not really part of the (mutable) model state.

src/simpleppl.jl
```diff
@@ -0,0 +1,231 @@
...
+Record the state of the model as a struct of NamedTuples, all
+sharing the same key values, namely, those of the model parameters.
+`value` should store the initial/current value of the parameters.
+`input` stores a tuple of inputs for a given node. `eval` are the
+anonymous functions associated with each node. These might typically
+be either deterministic values or some distribution, but could an 
+arbitrary julia program. `kind` is a tuple of symbols indicating
+whether the node is a logical or stochastic node.
+"""
+struct ModelState{T}
```
yebai, line 19: Maybe consider `DAGVarInfo` to stay consistent with the [`DynamicPPL` terminology ](https://github.com/TuringLang/DynamicPPL.jl/blob/1744ba7bbc9da5cd8e0d04ab66ba588030e30879/src/DynamicPPL.jl#L139)
```suggestion
struct GraphVarInfo{T} <: AbstractModelTrace
```

src/simpleppl.jl
```diff
@@ -0,0 +1,231 @@
...
+    kind::NamedTuple{T}
+end
+
+"""
+    DAG(inputs)
+
+Struct containing the adjacency matrix for a particular model and 
+the topologically ordered vertex list.
+"""
+struct DAG
```
yebai, line 32: Maybe consider `GraphInfo` for better consistency with `VarInfo` (i.e. model = varinfo + graphinfo):
```suggestion
struct GraphInfo
```

yebai (changes requested):

src/simpleppl.jl
yebai: @phipsgabler Since we have node attributes in `DAG`, we can store `value` and `input` in the same `values::NamedTuple`. If so, `GraphVarInfo` is basically the same as `SimpleVarInfo`. It motivates to perhaps moving `SimpleVarInfo` type definition into `AbstractPPL` in the future.  

https://github.com/TuringLang/DynamicPPL.jl/blob/master/src/simple_varinfo.jl#L129

src/simpleppl.jl
```diff
@@ -0,0 +1,231 @@
...
+DAG: 
+3×3 SparseArrays.SparseMatrixCSC{Bool, Int64} with 2 stored entries:
+⋅  ⋅  ⋅
+⋅  ⋅  ⋅
+1  1  ⋅
+```
+"""
+struct Model
+    ModelState::ModelState
+    DAG::DAG
```
yebai, line 72:
```suggestion
    gvi::GraphVarInfo
```

PavanChaggar:
@yebai I've merged `ModelState` and `DAG` together into one `GraphInfo` type. I've also just replaced the `Model` type with `GraphInfo`. Now when you call the `Model` constructor it returns a `GraphInfo` type. Is that ok or do you think it is better to have a dedicated `Model` type with field `g::GraphInfo`?

yebai:
@PavanChaggar thanks, let's still keep a dedicated `Model` type. We might add more fields to `Model` later. In addition, it is more consistent with the `Model` type in `DynamicPPL`, which we could bring closer in the future. In my mind, `model = CodeInfo + GraphInfo/VarInfo`, where `CodeInfo` is not yet introduced but could represent the model AST/IR.

yebai:
Let's also rename `simpleppl.jl` to `graphinfo.jl`.

phipsgabler:
I'd still like it to go into a separate submodule, so that it's clear that this is the part for one, concrete implementation.

src/AbstractPPL.jl
```diff
@@ -7,6 +7,8 @@ export VarName, getsym, getlens, inspace, subsumes, varname, vsym, @varname, @vs
 # Abstract model functions
 export AbstractProbabilisticProgram, condition, decondition, logdensityof, densityof
 
+# GraphInfo
+export GraphInfo, Model, dag, nodes
```
devmotion, line 11: The name `Model` and, to a smaller degree, `nodes` and `dag` seem to general to be exported.

PavanChaggar: No objections to getting rid of `nodes` and `dag`. I've put `GraphInfo` into a submodule `GraphPPL` that isn't exported, as per @phipsgabler suggestion. Does this mitigate concerns about directly exporting Model?

src/AbstractPPL.jl
```diff
@@ -17,4 +16,10 @@ include("abstractmodeltrace.jl")
 include("abstractprobprog.jl")
 include("deprecations.jl")
 
+# GraphInfo
+module GraphPPL
```
yebai, line 20:
```suggestion
module GraphInfo
```

src/AbstractPPL.jl
```diff
@@ -17,4 +16,10 @@ include("abstractmodeltrace.jl")
...
+    include("graphinfo.jl")    
+    export GraphInfo, Model, dag, nodes
```
yebai, line 22: For safety, let’s perhaps wait for a few more PRs before exporting anything. 
```suggestion
```

src/graphinfo.jl
```diff
@@ -0,0 +1,254 @@
...
+               y = (0.0, (μ, s2) -> MvNormal(μ, sqrt(s2)), :Stochastic)
+           )
+Nodes: 
+μ = (value = 1.0, input = (), eval = var"#6#9"(), kind = :Logical)
+s2 = (value = 0.0, input = (), eval = var"#5#8"(), kind = :Stochastic)
+y = (value = 0.0, input = (:μ, :s2), eval = var"#7#10"(), kind = :Stochastic)
+```
+"""
+
+struct Model{T}
```
yebai, line 58: @phipsgabler Would you like this to be a subtype of `AbstractProbabilisticProgram`?

src/graphinfo.jl
```diff
@@ -0,0 +1,254 @@
...
+    g::GraphInfo{T}
+end
+
+function Model(;kwargs...)
+    vals = getvals(NamedTuple(kwargs))
```
yebai, line 63: This function assumes the `kargs` are in a particular format. Perhaps consider adding a minimal set of sanity checking using assertations to avoid surprises?

src/graphinfo.jl
```diff
@@ -0,0 +1,254 @@
...
+
+
+"""
+    DAG(inputs)
+
+Function taking in a NamedTuple containing the inputs to each node 
+and returns the implied adjacency matrix and topologically ordered 
+vertex list.
+"""
+function DAG(inputs)
```
yebai, line 78: It’s unusual to use captical function names, perhaps
```suggestion
function dag(inputs)
```

src/graphinfo.jl
```diff
@@ -0,0 +1,254 @@
...
+
+"""
+    getvals(nt::NamedTuple{T})
+
+Takes in the arguments to Model(;kwargs...) as a NamedTuple and 
+reorders into a tuple of tuples each containing either of value, 
+input, eval and kind, as required by the GraphInfo type. 
+"""
+@generated function getvals(nt::NamedTuple{T}) where T
+    values = [:(nt[$i][$j]) for i in 1:length(T), j in 1:3]
```
yebai, line 95: Perhaps consider adding some sanity checking on `nt`, similar to those for `Model`?

yebai (commented):
Looks great - many thanks @PavanChaggar! I left some minor comments above.

src/graphinfo.jl
PavanChaggar: I've added checks in the `Model` constructor to make sure the order and types are as expected. If these are correct then `getvals` should work as expected.

src/graphinfo.jl
PavanChaggar: Just changed this

src/graphinfo.jl
PavanChaggar: added checks to make sure type and order are correct.

src/graphinfo.jl
PavanChaggar: @phipsgabler confirmed and I have done this.

yebai (approved):

Issue #46, opened 2022-01-03
Allow interpolation in `@varname`


phipsgabler:
Something like that should work:

```julia
julia> node = :a
:a

julia> @varname($node)
a
```

sunxd3:
Close as #54 merged.

PR #54, opened 2022-02-13
[Merged by Bors] - Allow name and property interpolation in varnames

phipsgabler:
Implements #46.

Most of this is dependent on my [PR in Setfield.jl](https://github.com/jw3126/Setfield.jl/pull/168) to allow interpolation of properties (so it works only if you `dev` that branch).

It's a bit of a hack though, since `parse_object_lens` behaves a bit weirdly when the innermost value of a lens chain is an interpolation:

```julia
julia> Setfield.parse_obj_lens(@q $name.a)
(:($(Expr(:escape, :_))), :((Setfield.compose)($(Expr(:escape, :name)), (Setfield.PropertyLens){:a}())))

julia> Setfield.parse_obj_lens(@q x.a)
(:($(Expr(:escape, :x))), :((Setfield.compose)((Setfield.PropertyLens){:a}())))
```

I'd be really nice to always get the latter form.  (`@q` is just an ad-hoc macro to construct expressions preserving the `:$`.)

@jw3126, could you maybe look at this and suggest any better alternatives? Either how to implement this nicely in AbstractPPL.jl, or improvements I can add to my PR.


jw3126:
> behaves a bit weirdly when the innermost value of a lens chain is an interpolation:

It is a feature not a bug :smile: . Until now interpolation meant lens interpolation and `$lens.a` was perfectly fine syntax to specify the composition of `lens` and `@lens _.a`. 

But now we have also "fieldname" interpolation and one might also expect "object" interpolation, which matches your desired parser output.

yebai (approved):
Many thanks, @phipsgabler for finalising this PR!

Issue #55, opened 2022-02-14
Move `AbstractContext` and `evaluate!!` to AbstractPPL


PavanChaggar:
Moving only the definitions of `AbstractContext` and `evaluate!!` to AbstractPPL. 

Relevant to DynamicPPL [#360](https://github.com/TuringLang/DynamicPPL.jl/pull/360).

yebai:
Fixed by #56

PR #56, opened 2022-02-19
[Merged by Bors] - moving AbstractContext and evaluate!! to AbstractPPL

PavanChaggar:
Addressing #55 and moving the definitions of `AbstractContext` and `evaluate!!` to AbstractPPL.

### DIFF
--- a/src/abstractprobprog.jl
+++ b/src/abstractprobprog.jl
@@ -60,3 +60,19 @@ m = decondition(condition(m, obs))
 should hold for generative models `m` and arbitrary `obs`.
 """
 function condition end
+
+""" 
+    AbstractContext
+
+Common base type for Contexts used for dispatching.
+"""
+abstract type AbstractContext end
+
+
+""" 
+    evaluate!!
+
+Sample from the `model` using the `sampler` with random number generator `rng` and the
+`context`, and store the sample and log joint probability in `varinfo`.
+"""
+function evaluate!! end
--- a/src/abstractprobprog.jl
+++ b/src/abstractprobprog.jl
@@ -64,7 +64,7 @@ function condition end
 """ 
     AbstractContext
 
-Common base type for Contexts used for dispatching.
+Common base type for evaluation contexts.
 """
 abstract type AbstractContext end
 
--- a/src/abstractprobprog.jl
+++ b/src/abstractprobprog.jl
@@ -60,10 +60,3 @@ m = decondition(condition(m, obs))
 should hold for generative models `m` and arbitrary `obs`.
 """
 function condition end
-
-""" 
-    AbstractContext
-
-Common base type for evaluation contexts.
-"""
-abstract type AbstractContext end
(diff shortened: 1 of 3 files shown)


yebai (approved):

src/abstractprobprog.jl
```diff
@@ -60,3 +60,19 @@ m = decondition(condition(m, obs))
...
+
+""" 
+    AbstractContext
+
+Common base type for Contexts used for dispatching.
+"""
+abstract type AbstractContext end
+
+
+"""
```
yebai, line 72: @PavanChaggar As @phipsgabler suggested, let's keep the `evaluate!` function in a separate file `evaluate.jl`.

src/abstractprobprog.jl
```diff
@@ -60,3 +60,19 @@ m = decondition(condition(m, obs))
...
+
+Common base type for Contexts used for dispatching.
+"""
+abstract type AbstractContext end
+
+
+""" 
+    evaluate!!
+
+Sample from the `model` using the `sampler` with random number generator `rng` and the
```
yebai, line 75: This doc string is no longer accurate. `evaluate!` is meant as a general API for various kinds of model operations, e.g. sampling from the prior, evaluating various log densities (prior, loglike and logjoint).

src/abstractprobprog.jl
```diff
@@ -60,3 +60,19 @@ m = decondition(condition(m, obs))
 should hold for generative models `m` and arbitrary `obs`.
 """
 function condition end
+
+""" 
+    AbstractContext
+
+Common base type for Contexts used for dispatching.
```
yebai, line 67:
```suggestion
Common base type for evaluation contexts.
```

yebai:
thanks, @PavanChaggar - I left some minor comments above. Can you also open a companion PR for DynamicPPL removing these types/functions?

src/abstractprobprog.jl
PavanChaggar: shall I keep `AbstractContext` in `abstractprobprog.jl`?

PavanChaggar:
companion pr for `DynamicPPL` is [here](https://github.com/TuringLang/DynamicPPL.jl/pull/382).

yebai (approved):

src/abstractprobprog.jl
yebai: That sounds like a better place - I've moved these lines.

devmotion:
Doesn't this create problems in DynamicPPL since both `AbstractContext` and `evaluate!!` are exported here and loaded in DynamicPPL?

yebai:
There is a companion PR for DynamicPPL, https://github.com/TuringLang/DynamicPPL.jl/pull/382. We will have to make a breaking release, then merge https://github.com/TuringLang/DynamicPPL.jl/pull/382 later. It is aiming to start the process of slowly unifying model operation APIs, which has been discussed in several places already.

Issue #51, opened 2022-02-07
Supporting mutating node values in `GraphInfo` 


yebai:
Our current `GraphInfo` stores `GraphInfo.value` in a `NamedTuple` which prevents mutation. 

https://github.com/TuringLang/AbstractPPL.jl/blob/469252618fd6ea85dc911b0999b4614fcca45f2f/src/graphinfo.jl#L25

One simple improvement is to use `Ref` to store values inside `NamedTupe`. It does not influence type stability and keep the overall design intact. For example: 

```julia
julia> Model(
               s2 = (0.0, () -> InverseGamma(2.0,3.0), :Stochastic), 
               μ = (1.0, () -> 1.0, :Logical), 
               y = (0.0, (μ, s2) -> MvNormal(μ, sqrt(s2)), :Stochastic)
           )
```

becomes 

```julia
julia> Model(
               s2 = (Ref(0.0), () -> InverseGamma(2.0,3.0), :Stochastic), 
               μ = (Ref(1.0), () -> 1.0, :Logical), 
               y = (Ref(0.0), (μ, s2) -> MvNormal(μ, sqrt(s2)), :Stochastic)
           )
```

Then we can do

```julia
# support mutation for stochastic variables
m[@varname s2] = 5.0 

# this should result in an error since we can't mutate logical variables (and hyperparameters)
m[@varname μ] = 2.0 
```

Note that we only need to change the internal model constructor and accessor functions and keep the user-facing `Model` constructor the same.

yebai:
Fixed by #57

PR #57, opened 2022-02-19
Remaking Model consturctor to turn values into Ref(values)

PavanChaggar:
Addressing #51 

Now the constructor will change the default values to `Ref(values)` so that they can be updated. I'll also add a convenience method for updating node values.


src/graphinfo.jl
```diff
@@ -217,6 +218,11 @@ function Base.getindex(m::Model, vn::VarName)
     return m.g[vn]
```
yebai, line 218: Do we need the following for dereferencing variables? 
```suggestion
    return m.g[vn][]
```

src/graphinfo.jl
```diff
@@ -63,7 +63,8 @@ function Model(;kwargs...)
     for (i, node) in enumerate(values(kwargs))
         @assert typeof(node) <: Tuple{Union{Array{Float64}, Float64}, Function, Symbol} "Check input order for node $(i) matches Tuple(value, function, kind)"
     end
-    vals = getvals(NamedTuple(kwargs))
+    vals = [getvals(NamedTuple(kwargs))...]
+    vals[1] = Tuple([Ref(Tuple(val)) for val in vals[1]])
```
yebai, line 67: Why we are wraping `val` inside a `Tuple`? 
```suggestion
    vals[1] = Tuple([Ref(val) for val in vals[1]])
```

PavanChaggar: I think otherwise `val` is a vector. Are we happy to keep `val` as a vector? I have no objections.

PavanChaggar: Probably easier to work with vectors for vals...

src/graphinfo.jl
PavanChaggar: I added a `setvalue!` function and got rid of the `setindex`.

yebai (approved):

yebai:
bors cancel

test/graphinfo.jl
```diff
@@ -36,8 +36,11 @@ A = sparse([0 0 0 0 0; 0 0 0 0 0; 0 0 0 0 0; 0 1 1 0 0; 1 0 0 1 0])
 
 # check the values from the NamedTuple match the values in the fields of GraphInfo
 vals = AbstractPPL.GraphPPL.getvals(model)
-for (i, field) in enumerate([:value, :eval, :kind])
-    @test eval( :( values(m.g.$field) == vals[$i] ) )
+for (i, v) in enumerate(vals[1])
+    @test values(m.g.value[i])[] == Ref(v)[]
```
yebai, line 40: @PavanChaggar can we modify `values` so that it internally dereferences `Ref` values so that we can still write `values(m.g.value[i]) == v`?

test/graphinfo.jl
```diff
@@ -57,6 +60,13 @@ single_parent_m = Model(μ = (1.0, () -> 3, :Logical), y = (1.0, (μ) -> MvNorma
 @test typeof(single_parent_m) == Model{(:μ, :y)}
 @test typeof(single_parent_m.g) == GraphInfo{(:μ, :y)}
 
+# test setindex
+
+@test_throws AssertionError setvalue!(m, @varname(s2), 0.0)
+@test_throws AssertionError setvalue!(m, @varname(s2), (1.0))
```
phipsgabler, line 66:
```suggestion
@test_throws AssertionError setvalue!(m, @varname(s2), (1.0,))
```
or
```suggestion
@test_throws AssertionError setvalue!(m, @varname(s2), 1.0)
```
which did you mean?

test/graphinfo.jl
```diff
@@ -57,6 +60,13 @@ single_parent_m = Model(μ = (1.0, () -> 3, :Logical), y = (1.0, (μ) -> MvNorma
...
+setvalue!(m, @varname(s2), [1.0, 1.0])
+@test m[@varname s2].value[] == [1.0,1.0]
```
phipsgabler, line 68: Let's have paired getters and setters, and completly avoid exposing the internals:

```suggestion
@test getvalue(m, @varname(s2)) == [1.0,1.0]
```

test/graphinfo.jl
```diff
@@ -36,8 +36,11 @@ A = sparse([0 0 0 0 0; 0 0 0 0 0; 0 0 0 0 0; 0 1 1 0 0; 1 0 0 1 0])
 
 # check the values from the NamedTuple match the values in the fields of GraphInfo
 vals = AbstractPPL.GraphPPL.getvals(model)
-for (i, field) in enumerate([:value, :eval, :kind])
-    @test eval( :( values(m.g.$field) == vals[$i] ) )
+for (i, v) in enumerate(vals[1])
+    @test values(m.g.value[i])[] == Ref(v)[]
+end
+for (i, field) in enumerate([:eval, :kind])
+    @test eval( :( values(m.g.$field) == vals[$i + 1] ) )
```
phipsgabler, line 43: No need for the heavy artillery here...

```suggestion
    @test values(getproperty(m.g, field)) == vals[i + 1] 
```

If ever this becomes necessary, prefer the macro version:

```julia
@eval(values(m.g.$field) == vals[$i + 1])
```

test/graphinfo.jl
phipsgabler: As said elsewhere, I'd completely forgo exposing the fields and have `getvalue`/`setvalue!` only.

src/graphinfo.jl
```diff
@@ -63,7 +63,8 @@ function Model(;kwargs...)
     for (i, node) in enumerate(values(kwargs))
         @assert typeof(node) <: Tuple{Union{Array{Float64}, Float64}, Function, Symbol} "Check input order for node $(i) matches Tuple(value, function, kind)"
```
phipsgabler, line 64: At least you didn't compare by `==`, but this is even more preferable:

```suggestion
        @assert node isa Tuple{Union{Array{Float64}, Float64}, Function, Symbol} "Check input order for node $(i) matches Tuple(value, function, kind)"
```

Also: Is the restriction to arrays and floats hard-coded anywhere? I'd strongly prefer allowing _any_ type of value.

phipsgabler (changes requested):

test/graphinfo.jl
PavanChaggar: The first. The comma hits me again...

yebai:
@PavanChaggar you might want to try `bors try` or `bors r` when you feel the PR is ready for CI / auto-merging.

yebai (approved):

yebai:
bors try

yebai:
bors try

PavanChaggar:
I've added some getters and setters for node values and refactored tests. Let me know if there's anything else to add.

yebai (approved):

yebai:
bors try

PavanChaggar:
bors try

yebai (approved):

yebai (approved):

PR #59, opened 2022-02-28
[Merged by Bors] - Bugfix for `evaluate!!` and `AbstractContext`

devmotion:
They are not defined and hence not actually exported currently. This explains also the test errors in the PRs in DynamicPPL and why the integration test did not complain when these functions were added to AbstractPPL.


yebai (approved):

yebai:
thanks, @devmotion - nice catch!

PR #63, opened 2022-04-08
 Implementing evaluation interface

PavanChaggar:
Working on implementing an interface for evaluating the model. This should include hooking into the `Random` API and `DensityInterface` API for density evaluation. I think this should be enough to get a simple example of MH sampling working with AbstractMCMC. I'll aim to implement contexts in another PR.


phipsgabler:
I'd rather prefer the MH example to live entirely in /test, or a separate /examples project. We don't need their dependencies for the interface, and it's also not core functionality.

Project.toml
```diff
@@ -8,11 +8,12 @@ version = "0.5.2"
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 DensityInterface = "b429d917-457f-4dbc-8f4c-0cc954292b1d"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
```
phipsgabler, line 11: We don't want to depend on Distributions in the main package. Rather, add it as a dependency to the Project.toml in test/.

Project.toml
```diff
@@ -8,11 +8,12 @@ version = "0.5.2"
...
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
-AbstractMCMC = "2, 3, 4"
```
phipsgabler, line 15: I think there still needs to be a bound?

src/graphinfo.jl
```diff
@@ -333,4 +394,156 @@ function Base.show(io::IO, m::Model)
...
+```jl-doctest
+julia> import AbstractPPL.GraphPPL: Model, rand!
+       using Distributions
+
+julia> using Random; Random.seed!(1234)
+TaskLocalRNG()
+
+julia> m = Model( s2 = (0.0, () -> InverseGamma(2.0,3.0), :Stochastic), 
+                                       μ = (1.0, () -> 1.0, :Logical), 
+                                       y = (0.0, (μ, s2) -> Normal(μ, sqrt(s2)), :Stochastic))
```
phipsgabler, line 415: The indentation of this looks off.

src/graphinfo.jl
```diff
@@ -333,4 +394,156 @@ function Base.show(io::IO, m::Model)
...
+MersenneTwister(1234)
+
+julia> import AbstractPPL.GraphPPL: Model, logdensityof
+[ Info: Precompiling AbstractPPL [7a57a42e-76ec-4ea3-a279-07e840d6d9cf]
+
+julia> using Distributions
+
+julia> m = Model( s2 = (1.0, () -> InverseGamma(2.0,3.0), :Stochastic), 
+                                              μ = (0.0, () -> 1.0, :Logical), 
+                                              y = (0.0, (μ, s2) -> Normal(μ, sqrt(s2)), :Stochastic))
```
phipsgabler, line 496: The indentation of this looks off.

src/graphinfo.jl
```diff
@@ -333,4 +394,156 @@ function Base.show(io::IO, m::Model)
...
+
+julia> import AbstractPPL.GraphPPL: Model, logdensityof, get_model_values
+[ Info: Precompiling AbstractPPL [7a57a42e-76ec-4ea3-a279-07e840d6d9cf]
+
+julia> using Distributions
+
+julia> m = Model( s2 = (1.0, () -> InverseGamma(2.0,3.0), :Stochastic), 
+                                              μ = (0.0, () -> 1.0, :Logical), 
+                                              y = (0.0, (μ, s2) -> Normal(μ, sqrt(s2)), :Stochastic))
+Nodes:
```
phipsgabler, line 529: Indentation

test/graphinfo/graphinfo.jl
```diff
@@ -61,19 +70,54 @@ for node in m
...
+
+for vn in keys(m)
+    mv = get_node_value(m, vn)
+    sv = get(sample, vn)
+    @test size(mv) == size(sv)
+    @test sv isa typeof(mv)
+end
+
+model1 = Model(μ = (3.0, () -> 3.0, :Logical),
+                    y = (0.0, (μ) -> Normal(μ, 1.0), :Stochastic))
```
phipsgabler, line 102: Indentation?

Project.toml
PavanChaggar: ah. Not sure how this got removed. I'll add it back in.

PavanChaggar:
I've now changed `rand` to use `deepcopy`. This doesn't work in Julia 1.7.2, but does work in 1.6.6 and I think is fixed in 1.7.3. How shall we bound Julia versions accordingly?

phipsgabler:
I _think_ it's `"1.6, 1.7.3"`, but ask first on Slack, please :P

PavanChaggar:
@phipsgabler according to the slack wizards, `julia = "~1.6.6, 1.7.3` is the compat we need, which I've just added to the project.toml

yebai (approved):
Many thanks, @PavanChaggar - it looks nice. Some function names are a bit verbose (e.g. `get_model_values`/`get_model_ref_values` but that's probably not very important and can be improved later if needed.

yebai:
bors try

Project.toml
```diff
@@ -16,4 +16,4 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 AbstractMCMC = "2, 3, 4"
 DensityInterface = "0.4"
 Setfield = "0.7.1, 0.8"
-julia = "1.6"
+julia = "~1.6.6, 1.7.3"
```
phipsgabler, line 19: Hei @devmotion, you are more versioning wizard than I -- basicallywe want to disallow 1.7.x below patch version 3 here, as deepcopy is buggy in those (#66). Is this a good bound for that?

yebai: Probably fine to ignore this issue since we haven't even exported the `GraphPPL` module yet.

PavanChaggar:
bors try

yebai (approved):

PR #75, opened 2022-12-28
[Merged by Bors] - Fix concretization for `varname`

torfjelde:
Use `Setfield.need_dynamic_lens` as the default value for `concretize` in `varname`. This way we preserve pre-0.6 behavior, thus not make the 0.6-release "less" breaking.

Ref: https://github.com/TuringLang/DynamicPPL.jl/pull/440

### DIFF
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -579,7 +579,8 @@ macro varname(expr::Union{Expr,Symbol}, concretize::Bool=false)
     return varname(expr, concretize)
 end
 
-varname(sym::Symbol, concretize=false) = :($(AbstractPPL.VarName){$(QuoteNode(sym))}())
+varname(sym::Symbol) = :($(AbstractPPL.VarName){$(QuoteNode(sym))}())
+varname(sym::Symbol, _) = varname(sym)
 function varname(expr::Expr, concretize=Setfield.need_dynamic_lens(expr))
     if Meta.isexpr(expr, :ref) || Meta.isexpr(expr, :.)
         # Split into object/base symbol and lens.
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -580,7 +580,7 @@ macro varname(expr::Union{Expr,Symbol}, concretize::Bool=false)
 end
 
 varname(sym::Symbol, concretize=false) = :($(AbstractPPL.VarName){$(QuoteNode(sym))}())
-function varname(expr::Expr, concretize=false)
+function varname(expr::Expr, concretize=Setfield.need_dynamic_lens(expr))
     if Meta.isexpr(expr, :ref) || Meta.isexpr(expr, :.)
         # Split into object/base symbol and lens.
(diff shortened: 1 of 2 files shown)


src/varname.jl
```diff
@@ -575,12 +575,13 @@ julia> x = (a = [(b = rand(2), )], ); getlens(@varname(x.a[1].b[end], true))
...
     Julia 1.5.
 """
-macro varname(expr::Union{Expr,Symbol}, concretize::Bool=false)
+macro varname(expr::Union{Expr,Symbol}, concretize::Bool=Setfield.need_dynamic_lens(expr))
     return varname(expr, concretize)
 end
 
-varname(sym::Symbol, concretize=false) = :($(AbstractPPL.VarName){$(QuoteNode(sym))}())
-function varname(expr::Expr, concretize=false)
+varname(sym::Symbol) = :($(AbstractPPL.VarName){$(QuoteNode(sym))}())
```
yebai, line 582: Presumably, these two `varname` are for backward compatibility. If so, maybe add a comment, so we don't forget.

yebai (approved):

yebai:
bors try

src/varname.jl
torfjelde: Nah, it has nothing to do with backwards compat. It's just that IMO this makes it clearer that `varname(::Symbol, ...)` does not at all depend on `concretize` vs. the previous impl which has `concretize=false` for `::Symbol`.

I can revert the change if you disagree:)

yebai: Oh, that makes sense. I am happy to keep them but perhaps add a comment to clarify your point.

torfjelde:
This won't fix the DPPL tests alone. The PR causing these fundamentally changes how we treat `Colon`, hence we need to change DPPL to accommodate these.

src/varname.jl
```diff
@@ -414,6 +414,8 @@ end
 
 subsumes_index(i::Colon, ::Colon) = error("Colons cannot be subsumed")
```
torfjelde, line 415: This line causes downstream issues @yebai

torfjelde: It affects every model which uses `:` for indexing.

torfjelde: Because we then try to call `subsumes` on it when checking.

yebai (approved):

src/varname.jl
torfjelde: I might have a fix

yebai: I remember these issues from the latent Dirichlet allocation model and MCMCChains; IMO, we should always replace `:` with `1:N` where N is a constant.

torfjelde: Yeah, but this issue goes a bit deeper than that.

The PR causing these issues were trying to address things like OffsetArrays.jl not working properly, etc. So I think we're now also in need of concretization stuff like `x[2:3]`, etc. I'm trying to figure this out.

PR #80, opened 2023-02-25
Backport 3-arg `rand` to AbstractPPL 0.5.x

sunxd3:
`rand` defaults have been added since AbstractPPL v0.6.2 (#79), backporting to v0.5 for compatibility.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -12,6 +12,7 @@ on:
   pull_request:
     branches:
       - main
+      - releases-0.5.x
 
 jobs:
   test:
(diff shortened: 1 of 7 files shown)


devmotion:
For future backports, a common approach (that we use e.g. also in SpecialFunctions) is to cherry-pick the relevant commit. See e.g. https://github.com/JuliaLang/julia/blob/a77ba79855f400420ccaa09c153ecec2784e386e/doc/src/devdocs/build/distributing.md#cherry-picking-commits for an explanation. The advantage is that one can then resolve conflicts similar to merges and git copies information such as the original commit message and contributors.

sunxd3:
The changes were done by local `git cherry-pick`, somehow the original commit information was lost still, I'll verify next time.

PR #81, opened 2023-02-25
Add StatsBase.predict to the interface

sethaxen:
As suggested in https://github.com/TuringLang/DynamicPPL.jl/issues/466#issuecomment-1436670214, this PR adds `StatsBase.predict` to the API with a default implementation.

### DIFF
--- a/src/abstractprobprog.jl
+++ b/src/abstractprobprog.jl
@@ -1,6 +1,7 @@
 using AbstractMCMC
 using DensityInterface
 using Random
+using StatsBase
 
 
 """
@@ -79,4 +80,30 @@ function Base.rand(::Type{T}, model::AbstractProbabilisticProgram) where {T}
 end
 function Base.rand(model::AbstractProbabilisticProgram)
     return rand(Random.default_rng(), NamedTuple, model)
+end
+
+"""
+    predict(
+        [rng::AbstractRNG=Random.default_rng(),]
+        [T=NamedTuple,]
+        model::AbstractProbabilisticProgram,
+        params,
+    ) -> T
+
+Draw a sample from the joint distribution specified by `model` conditioned on the values in
+`params`.
+
+The sample will be returned as format specified by `T`.
+"""
+function StatsBase.predict(rand::AbstractRNG, T::Type, model::AbstractProbabilisticProgram, params)
+    return rand(rng, T, condition(model, params))
+end
+function StatsBase.predict(T::Type, model::AbstractProbabilisticProgram, params)
+    return StatsBase.predict(Random.default_rng(), T, model, params)
+end
+function StatsBase.predict(model::AbstractProbabilisticProgram, params)
+    return StatsBase.predict(NamedTuple, model, params)
+end
+function StatsBase.predict(rng::AbstractRNG, params)
+    return StatsBase.predict(rng, NamedTuple, model, params)
 end
--- a/Project.toml
+++ b/Project.toml
@@ -11,9 +11,11 @@ DensityInterface = "b429d917-457f-4dbc-8f4c-0cc954292b1d"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
 AbstractMCMC = "2, 3, 4"
 DensityInterface = "0.4"
 Setfield = "0.8.2, 1"
+StatsBase = "0.32, 0.33"
 julia = "~1.6.6, 1.7.3"


src/abstractprobprog.jl
```diff
@@ -80,3 +81,29 @@ end
...
+        params,
+    ) -> T
+
+Draw a sample from the joint distribution specified by `model` conditioned on the values in
+`params`.
+
+The sample will be returned as format specified by `T`.
+"""
+function StatsBase.predict(rand::AbstractRNG, T::Type, model::AbstractProbabilisticProgram, params)
+    return rand(rng, T, condition(model, params))
```
sethaxen, line 99: The docstring for `condition` would technically limit `params` to observations, for which this definitely does the wrong thing. Should `condition` be generalized?

src/abstractprobprog.jl
```diff
@@ -80,3 +81,29 @@ end
...
+
+"""
+    predict(
+        [rng::AbstractRNG=Random.default_rng(),]
+        [T=NamedTuple,]
+        model::AbstractProbabilisticProgram,
+        params,
+    ) -> T
+
+Draw a sample from the joint distribution specified by `model` conditioned on the values in
```
sethaxen, line 93: "condition" isn't quite right. If the model has a probabilistic graph `x -> y -> z`, and `params` contains `y`, this will sample `x ~ p(x)` and `z ~ p(z | y)`, when conditioning would give `x ~ p(x | y)` and `z ~ p(z | y)`

sethaxen: I guess we actually want a `do` operator here instead of `condition`.

sethaxen:
Bump, maybe @devmotion or @torfjelde?

sethaxen:
Bump again.

src/abstractprobprog.jl
torfjelde: This is a good point.

> I guess we actually want a do operator here instead of condition.

But doesn't `do` cut the graph too, and so it would also remove the `x` completely?

src/abstractprobprog.jl
torfjelde: > The docstring for condition would technically limit params to observations

What do you mean by this?

torfjelde (commented):
Cheers for the bump; had missed this!

It's worth noting that DPPL is still not compatible wtih AbstractPPL@0.6 so we might also want to add this to AbstractPPL@0.5.

Furthermore, I'm slightly worried about the state of AbstractPPL atm; it's not clear if anyone has any ownership of the package atm, and IMO it's objectives are a bit all over the place.
I'd personally be happy to go against what was originally suggested in https://github.com/TuringLang/DynamicPPL.jl/issues/466#issuecomment-1436670214 and just putting this directly in DPPL.

_Or_ we need to start giving AbstractPPL some love :confused:

src/abstractprobprog.jl
sethaxen: > But doesn't `do` cut the graph too, and so it would also remove the `x` completely?

As I understand it, `do(y=something)` only deletes the edges going into `y` and sets `y' to the specified value. The rest of the graph is unchanged; in particular, no nodes are deleted. So `x` would be kept.

src/abstractprobprog.jl
sethaxen: The docstring for `condition` reads:
> Condition the generative model `model` on some observed data

So as documented it is only meant for conditioning on observations, or equivalently, leaves in a DAG. But here `predict` wants to fix everything _but_ the leaves in the DAG (or perhaps more generally, a subgraph that contains a leaf).

src/abstractprobprog.jl
torfjelde: Ah yes that makes sense.

Okay so we basically want `do`to be like `condition` but without including the accumulation of it's (log) prob?

sunxd3:
I think @sethaxen might be preoccupied, so I am taking over. Let me know if this is bad.

sunxd3:
apologies for the ping, this might not be ready yet, but maybe time to take a look and start some new discussions

torfjelde:
I guess the one question is whether we should perform `decondition` in `predict` or not?

sunxd3:
> I guess the one question is whether we should perform decondition in predict or not?

just saw the comment, sorry. I thought about the same thing, but unsure what is the right thing to do. The issue for me is that the dimension of the prediction might not match the dimension of the data.

How about we don't give a default implementation right now? 

To clarify, the default implementations for the optional arguments should be included, but not

```julia
function StatsBase.predict(rng::AbstractRNG, T::Type, model::AbstractProbabilisticProgram, params)
    return rand(rng, T, fix(model, params))
end
```

src/abstractprobprog.jl
```diff
@@ -121,3 +122,28 @@ end
...
+    ) -> T
+
+Draw a sample from the predictive distribution specified by `model` with its parameters fixed to `params`.
+
+The sample will be returned as format specified by `T`.
+"""
+function StatsBase.predict(rng::AbstractRNG, T::Type, model::AbstractProbabilisticProgram, params)
+    return rand(rng, T, fix(model, params))
+end
+function StatsBase.predict(T::Type, model::AbstractProbabilisticProgram, params)
```
sunxd3, line 141: @torfjelde @yebai @penelopeysm do you think type `T` still a good idea?

I think it's probably okay for the function to be a bit under-speced now, so DynamicPPL and JuliaBUGS and others can decide what type to return.

sunxd3: I think when Seth added this, it was modeled after the `rand` interface
https://github.com/TuringLang/AbstractPPL.jl/blob/b72a96368dc247549b9bbe10e13282ccd003517c/src/abstractprobprog.jl#L108-L124

I am for just having a simple `predict` now, or at most with `rng`, but not output type `T`.

Moreover, should we slim down the `rand` interface also, as this is going to be a breaking release.

penelopeysm: If we want DynamicPPL interface to have a type argument `T`, it makes sense for this one to have `T` as well. Otherwise I don't really see the point of having some interface here and then giving it a different signature in DynamicPPL.

penelopeysm (approved):
Personally, I'm happy to merge as is.

This PR doesn't bump version. However, the version was previously bumped in #109 to 0.10.0 and we didn't make a release, so I think it's fine to merge this and then release 0.10.0.

src/abstractprobprog.jl
sunxd3: Agree with this, my comments only reflects that we don't have T argument right now if I recall correctly

penelopeysm: Ohhh, I see! For a general interface, though, if we don't specify `T` then we have to choose a privileged output type (like NamedTuple) right? Otherwise if it can return anything then it's not super useful either.

What do you think of something like this:

```julia
# default_return_type(model) specifies the default type returned by
# rand([rng, ]model) and predict([rng, ]model, params)
function default_return_type end

# Then we can have rand like this
function Base.rand(
    rng::Random.AbstractRNG = Random.default_rng(),
    ::Type{T} = default_return_type(model),
    model::AbstractProbabilisticProgram)
)
     # whatever implementation
end

# And predict like this
function StatsBase.predict(
    rng::Random.AbstractRNG = Random.default_rng(),
    ::Type{T} = default_return_type(model),
    model::AbstractProbabilisticProgram),
    params
)
     # whatever implementation
end 
```

penelopeysm: Basically, I don't feel super comfortable only being able to return NamedTuple. I think that's fine for a downstream package like DynamicPPL to choose, but enforcing it at this level might be too limiting. I also know I'm possibly overcomplicating things, sorry 😄

sunxd3: @penelopeysm really sorry for missing the previous reply, your comments makes perfect sense to me. I think it's a good idea to let user decide what type to return. My view is: it might also not be a bad idea to let PPL packages that implements `AbstractPPL` decide what type to return, or return the same type as `params`.

I am just a bit uneasy to have a default return type that we don't always support (`OrderedDict` would be better, but also not perfect).

Issue #82, opened 2023-03-27
Is the graph-based interface used anywhere?


torfjelde:
AFAIK the graph-based interface in AbstractPPL.jl isn't actually used for anything?

Hence, are people opposed to just removing the code related to it?

I'm slightly worried it's just laying there dorment, without anyone using nor working on it. I was reminded of this because of an versioning-issue someone ran into on Julia 1.7.2 because we're restricted to Julia 1.7.3 and upwards on AbstractPPL@0.[5-6].2, which is a change that traces back to a related PR (https://github.com/TuringLang/AbstractPPL.jl/pull/63).

PR #85, opened 2023-10-24
Remove `graphinfo.jl`

sunxd3:
Per #82

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "7a57a42e-76ec-4ea3-a279-07e840d6d9cf"
 keywords = ["probablistic programming"]
 license = "MIT"
 desc = "Common interfaces for probabilistic programming"
-version = "0.7.1"
+version = "0.7.0"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"


Project.toml
```diff
@@ -3,7 +3,7 @@ uuid = "7a57a42e-76ec-4ea3-a279-07e840d6d9cf"
 keywords = ["probablistic programming"]
 license = "MIT"
 desc = "Common interfaces for probabilistic programming"
-version = "0.6.3"
+version = "0.6.4"
```
torfjelde, line 6: Is this not breaking?

torfjelde (changes requested):

Project.toml
```diff
@@ -3,7 +3,7 @@ uuid = "7a57a42e-76ec-4ea3-a279-07e840d6d9cf"
 keywords = ["probablistic programming"]
 license = "MIT"
 desc = "Common interfaces for probabilistic programming"
-version = "0.6.3"
+version = "0.7.1"
```
torfjelde, line 6: If this is the first breaking release, then this should be
```suggestion
version = "0.7.0"
```

torfjelde (changes requested):

PR #88, opened 2023-11-08
Add `AbstractMCMC`v5.0 to Project.toml

sunxd3:
https://github.com/TuringLang/DynamicPPL.jl/pull/551


devmotion:
Backporting?

devmotion:
It seems AbstractPPL is not affected by the breaking changes in AbstractMCMC and does not export any of its functionality. So it should be fine to tag a non-breaking release IMO.

PR #89, opened 2023-11-08
Bump `AbstractMCMC` version and backporting to v0.6

sunxd3:


devmotion:
It doesn't matter here I think but ideally commits are backported with `git cherry-pick`. Then the backport commit is linked to the original commit on the master branch.

PR #91, opened 2024-03-09
Move to `Accessors.jl` from `Setfield.jl`

sunxd3:
## Changes from `Setfield` to `Accessors`
* `Accessors` is based on [`ComposeFunction`](https://docs.julialang.org/en/v1/base/base/#Base.ComposedFunction), which means all the `optic`s are either function or functor. Also, no `get(obj, optic)` interface, but `optic(obj)`.
* `Lens` abstract type doesn't exist in `Accessors`, `IdentityLens` is just `identity`, `ComposedLens` is now `ComposedOptic` (alias for `ComposedFunction`)
* `Inner` and `Outer` are interchanged, `l1 ∘ l2` in `Setfield` is `l2 ∘ l1` in `Accessors`. `Accessors` adopted `⨟`(`\bbsemi` "opposite compose") from `CompositionBase`: `l1 ∘ l2` in `Setfield` is `l1 ⨟ l2` in `Accessors`.

## Unsure
* The equivalence between `VarNames` are changed, before `@varname(y[:], true)` and `@varname(y[1:100]` (where `y = zeros(10,10)`) are considered to be `==`. But with `Accesssors`, two `IndexLens` are generally not equivalent even when they have the same field because they are also functors.
* ~~`Accessor.parse_obj_optic` and `Setfield.parse_obj_lens` behave differently when deal with interpolation. I have tried and failed to maintain the current interface with `Accessor.parse_obj_optic`, so I am keeping `Setfield.parse_obj_lens` (and `Setfield` as a dependency), while replace the `Lens` with optic in the result of `Setfield.parse_obj_lens`.~~ 
* Optics are functions or functors, how would this affect type stability?

### DIFF
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -630,10 +630,10 @@ function varname(expr::Expr, concretize=Accessors.need_dynamic_optic(expr))
         sym = drop_escape(sym_escaped)
 
         # This is to handle interpolated heads -- Setfield treats them differently:
-        # julia> _parse_obj_optics(@q $name.a)
-        # (:($(Expr(:escape, :_))), :((Setfield.compose)($(Expr(:escape, :name)), (Setfield.PropertyLens){:a}())))
-        # julia> _parse_obj_optics(@q x.a)
-        # (:($(Expr(:escape, :x))), :((Setfield.compose)((Setfield.PropertyLens){:a}())))
+        # julia>  AbstractPPL._parse_obj_optics(Meta.parse("\$name.a"))
+        # (:($(Expr(:escape, :_))), (:($(Expr(:escape, :name))), :((PropertyLens){:a}())))
+        # julia> AbstractPPL._parse_obj_optic(:(x.a))
+        # (:($(Expr(:escape, :x))), :(Accessors.opticcompose((PropertyLens){:a}())))
         if sym != :_
             sym = QuoteNode(sym)
         else


src/varname.jl
```diff
@@ -1,70 +1,71 @@
...
 
 """
-    is_static_lens(l::Lens)
-
-Return `true` if `l` does not require runtime information to be resolved.
+    is_static_optic(l)
 
-In particular it returns `false` for `Setfield.DynamicLens` and `Setfield.FunctionLens`.
+Return `true` if `l` is one or a composition of `identity`, `PropertyLens`, and `IndexLens`.
 """
```
torfjelde, line 51: Is there a reason why the docstring was changed here?

src/varname.jl
```diff
@@ -104,78 +105,83 @@ julia> getsym(@varname(y))
...
+    print(io, shortstr)
 end
-_print_application(io::IO, l::IndexLens) =
-    print(io, "[", join(map(prettify_index, l.indices), ","), "]")
-# This is a bit weird but whatever. We're almost always going to
-# `concretize` anyways.
-_print_application(io::IO, l::DynamicIndexLens) = print(io, l, "(_)")
+# TODO: type piracy https://github.com/JuliaObjects/Accessors.jl/blob/01528a81fdf17c07436e1f3d99119d3f635e4c26/src/sugar.jl#L495
+Accessors._shortstring(prev, o::IndexLens) = "$prev[$(join(map(prettify_index, o.indices), ", "))]"
+Accessors._shortstring(prev, o::typeof(identity)) = "$prev"
```
torfjelde, line 176: Why do we need this? Maybe add a comment as to why we're overloading this stuff.

And do what we did before: if you're going to overload behavior of something we don't own, then define a new `_shortstring` method which defaults to `Accessors._shortstring` but can be further overloaded without piracy.

That is,
```julia
_shortstring(prev, o) = Accessors._shortstring(prev, o)
# Specializations
...
```
and then call this in `_print_application` instead.
and then

src/varname.jl
```diff
@@ -266,28 +272,28 @@ Currently _not_ supported are:
   - Trailing ones: `x[2, 1]` does not subsume `x[2]` for a vector `x`
 """
 function subsumes(u::VarName, v::VarName)
-    return getsym(u) == getsym(v) && subsumes(u.lens, v.lens)
+    return getsym(u) == getsym(v) && subsumes(u.optic, v.optic)
```
torfjelde, line 275:
```suggestion
    return getsym(u) == getsym(v) && subsumes(getoptic(optic), getoptic(v))
```

This should have been done originally for the lens stuff, but let's just fix it now:)

src/varname.jl
```diff
@@ -104,78 +105,83 @@ julia> getsym(@varname(y))
...
 
 # Allow compositions with lenses.
-function Base.:∘(vn::VarName{sym,<:Lens}, lens::Lens) where {sym}
-    return VarName{sym}(getlens(vn) ∘ lens)
+function Base.:∘(vn::VarName{sym,T}, optic) where {sym,T}
+    return VarName{sym}(getoptic(vn) ⨟ optic)
+end
+function Base.:∘(vn::VarName{sym,typeof(identity)}, optic) where {sym}
+    return VarName{sym}(optic)
 end
```
torfjelde, line 154: Hmm, I'm a bit uncertain about this.

We implemented this originally to replicate the composition behavior of Setfield.jl, but now this reversed and so it seems a bit strange to not also reverse this here?

torfjelde (changes requested):
> The equivalence between VarNames are changed, before @varname(y[:], true) and @varname(y[1:100] (where y = zeros(10,10)) are considered to be ==. But with Accesssors, two IndexLens are generally not equivalent even when they have the same field because they are also functors.

I don't get this. Why does it matter that it's a functor?

From what I can tell it seems the issue is that `@varname(y[1:100])` results in the indices field holding the range `1:100`, not `ConcretizedSlice(Base.OneTo(100))`, which is what `@varname(y[:], true)` results in?

You're saying this was _not_ the case before?

> Optics are functions or functors, how would this affect type stability?

Maybe add some tests with `@inferred`?

Though for the subset of cases where care about, I'd be surprised if this would actually have any implications for type-stability.

The only worry is whether the fact that subtypes of `Function` will in certain cases not be specialized on, but `Base.ComposedFunction` does not subtype `Function` + from what I can tell, none of the optics subtype `Function` either, so we should be good.

src/varname.jl
```diff
@@ -1,70 +1,70 @@
...
-A variable identifier for a symbol `sym` and lens `lens`.
+A variable identifier for a symbol `sym` and optic `optic`.
 
 The Julia variable in the model corresponding to `sym` can refer to a single value or to a
 hierarchical array structure of univariate, multivariate or matrix variables. The field `lens`
 stores the indices requires to access the random variable from the Julia variable indicated by `sym`
-as a tuple of tuples. Each element of the tuple thereby contains the indices of one lens
-operation.
+as a tuple of tuples. Each element of the tuple thereby contains the indices of one optic
+operation.
```
torfjelde, line 16:
```suggestion
operation.
```

torfjelde (changes requested):
> The interpolation is weaker compared to before. It is mainly attributed to Accessors.parse_obj_optic. Now more than one interpolated symbol (e.g. @varname($n1[$n2+1])), and interpolation of the LHS of a . expr (e.g. $name.a[1]) are disallowed.

Woaah this seems not great; why do we feel the need to make this change?

src/varname.jl
torfjelde: The thing is, we are making `VarName` act as, what was before, a `PropertyLens`, in which case it _does_ make sense to consider the composition IMO. We are still doing this, no?

torfjelde:
>  By saying "they are functor", I am just making a guess why Accessors doesn't have this == defined as Setfield does: it might be related to the semantic meaning of functions equivalence.

Might be worth raising an issue and asking about this, as IIUC it should be okay to define `==` for some of the optics, e.g the getindex one.

torfjelde:
Is there a reason why we can't just copy-paste the macro? :shrug:

torfjelde (approved):
Great stuff @sunxd3 :)

PR #92, opened 2024-03-31
Remove MacroTools

yebai:
Related: https://github.com/JuliaRegistries/General/pull/103811#issuecomment-2025838678

### DIFF
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -673,10 +673,9 @@ end
 # Accessors doesn't have the same support for interpolation
 # so this function is copied and altered from `Setfield._parse_obj_lens`
 function _parse_obj_optics(ex)
-    if ex isa Expr && ex.head == :$
-        @assert length(ex.args) == 1
+    if Meta.isexpr(ex, :$, 1)
         return esc(:_), (esc(ex.args[1]),)
-    elseif Meta.isexpr(ex, :ref)
+    elseif Meta.isexpr(ex, :ref) && !isempty(ex.args)
         front, indices... = ex.args
         obj, frontoptics = _parse_obj_optics(front)
         if any(Accessors.need_dynamic_optic, indices)
@@ -689,13 +688,13 @@ function _parse_obj_optics(ex)
             index = esc(Expr(:tuple, indices...))
             optics = :($(Accessors.IndexLens)($index))
         end
-    elseif Meta.isexpr(ex, :.)
+    elseif Meta.isexpr(ex, :., 2)
         front = ex.args[1]
         property = ex.args[2].value # ex.args[2] is a QuoteNode
         obj, frontoptics = _parse_obj_optics(front)
         if property isa Union{Symbol,String}
             optics = :($(Accessors.PropertyLens){$(QuoteNode(property))}())
-        elseif property isa Expr && property.head == :$
+        elseif Meta.isexpr(property, :$, 1)
             optics = :($(Accessors.PropertyLens){$(esc(property.args[1]))}())
         else
             throw(ArgumentError(
--- a/test/varname.jl
+++ b/test/varname.jl
@@ -89,7 +89,6 @@ end
         @test collect(get(B, @varname(B[1, :], true))) == collect(get(B, @varname(B[1, -4:5])))
 
     end
-    
     @testset "type stability" begin
         @inferred VarName{:a}()
         @inferred VarName{:a}(IndexLens(1))


src/varname.jl
```diff
@@ -1,6 +1,6 @@
 using Accessors
 using Accessors: ComposedOptic, PropertyLens, IndexLens, DynamicIndexLens
-using MacroTools
+using Accessors.MacroTools: @capture
```
devmotion, line 3: Accessors may remove the dependency on MacroTools at any time in a non-breaking release. Additionally, it may switch to a new breaking release of MacroTools and we can't prevent this from breaking `@capture`. So IMO we should either add a compat entry for MacroTools (probably easiest short term solution) or rewrite `@capture` and remove MacroTools completely (probably best long term approach).

src/varname.jl
```diff
@@ -671,30 +670,14 @@ function _parse_obj_optic(ex)
...
-    end
-end
-
+# Accessors doesn't have the same support for interpolation
+# so this function is copied and altered from `Setfield._parse_obj_lens`
 function _parse_obj_optics(ex)
-    if @capture(ex, ∘(opticsexprs__))
-        return _parse_obj_optics_composite(opticsexprs)
-    elseif is_interpolation(ex)
+    if ex isa Expr && ex.head == :$
```
devmotion, line 676:
```suggestion
    if Meta.isexpr(ex, :$, 1)
```

src/varname.jl
```diff
@@ -671,30 +670,14 @@ function _parse_obj_optic(ex)
...
         @assert length(ex.args) == 1
```
devmotion, line 677: Not needed when the `Meta.isexpr` suggestion above is applied.
```suggestion
```

src/varname.jl
```diff
@@ -671,30 +670,14 @@ function _parse_obj_optic(ex)
...
         return esc(:_), (esc(ex.args[1]),)
-    elseif @capture(ex, front_[indices__])
+    elseif Meta.isexpr(ex, :ref)
```
devmotion, line 679: The next line will fail if there is not at least one argument:
```suggestion
    elseif Meta.isexpr(ex, :ref) && !isempty(ex.args)
```

src/varname.jl
```diff
@@ -706,11 +689,13 @@ function _parse_obj_optics(ex)
             index = esc(Expr(:tuple, indices...))
             optics = :($(Accessors.IndexLens)($index))
         end
-    elseif @capture(ex, front_.property_)
+    elseif Meta.isexpr(ex, :.)
```
devmotion, line 692:
```suggestion
    elseif Meta.isexpr(ex, :., 2)
```

src/varname.jl
```diff
@@ -706,11 +689,13 @@ function _parse_obj_optics(ex)
...
+        front = ex.args[1]
+        property = ex.args[2].value # ex.args[2] is a QuoteNode
         obj, frontoptics = _parse_obj_optics(front)
         if property isa Union{Symbol,String}
             optics = :($(Accessors.PropertyLens){$(QuoteNode(property))}())
-        elseif is_interpolation(property)
+        elseif property isa Expr && property.head == :$
```
devmotion, line 698:
```suggestion
        elseif Meta.isexpr(property, :$, 1)
```

test/varname.jl
```diff
@@ -89,7 +89,7 @@ end
         @test collect(get(B, @varname(B[1, :], true))) == collect(get(B, @varname(B[1, -4:5])))
 
     end
-
+
```
devmotion, line 92:
```suggestion

```

src/varname.jl
```diff
@@ -671,30 +670,14 @@ function _parse_obj_optic(ex)
...
-        end
-        return obj, (outermostlens, innerlenses...)
-    end
-end
-
+# Accessors doesn't have the same support for interpolation
+# so this function is copied and altered from `Setfield._parse_obj_lens`
 function _parse_obj_optics(ex)
-    if @capture(ex, ∘(opticsexprs__))
-        return _parse_obj_optics_composite(opticsexprs)
```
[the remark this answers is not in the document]
torfjelde: I don't fully understand this comment :confused: 

Are you saying that by removing this we no longer support something like `@varname(x.a[1])`?

torfjelde (changes requested):
Need some clarification regarding the change from @sunxd3 here, but otherwise looks good. So will approve if it turns out to be nothing:)

torfjelde (approved):
LGTM! Thanks @sunxd3 :)

PR #95, opened 2024-04-09
Bring back `∘` function between varname and optic

sunxd3:
The `∘` function was removed in #91, this PR puts it back, but adjusts the order to be consistent with `Accessors`.
Also added `opcompose` (_op_posite compose, alias of `⨟`).

### DIFF
--- a/src/varname.jl
+++ b/src/varname.jl
@@ -1,6 +1,5 @@
 using Accessors
 using Accessors: ComposedOptic, PropertyLens, IndexLens, DynamicIndexLens
-import Accessors: opcompose, ⨟
 
 const ALLOWED_OPTICS = Union{typeof(identity),PropertyLens,IndexLens,ComposedOptic}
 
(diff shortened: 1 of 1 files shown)


src/varname.jl
```diff
@@ -1,5 +1,6 @@
 using Accessors
 using Accessors: ComposedOptic, PropertyLens, IndexLens, DynamicIndexLens
+import Accessors: opcompose, ⨟
```
devmotion, line 3: This should be unnecessary, Accessors reexports them (they are defined in CompositionsBase). They're aliases, so one of them would be sufficient. Generally, though, I'd avoid `import` since it makes it easier to accidentally extend methods.
```suggestion
```

src/varname.jl
```diff
@@ -148,6 +149,13 @@ function Accessors.set(obj, vn::VarName{sym}, value) where {sym}
     return Accessors.set(obj, PropertyLens{sym}() ⨟ getoptic(vn), value)
 end
 
+# Allow compositions with optic.
+function Base.:∘(optic::ALLOWED_OPTICS, vn::VarName{sym,<:ALLOWED_OPTICS}) where {sym}
+    return VarName{sym}(optic ∘ getoptic(vn))
+end
+function Accessors.opcompose(vn::VarName{sym,<:ALLOWED_OPTICS}, optic::ALLOWED_OPTICS) where {sym}
```
devmotion, line 156: This can be removed AFAICT, the fallback in https://github.com/JuliaFunctional/CompositionsBase.jl/blob/9c80024d40108ca873bf6351a14246c0165665d9/src/CompositionsBase.jl#L21 should be sufficient.
```suggestion
```

src/varname.jl
```diff
@@ -148,6 +149,13 @@ function Accessors.set(obj, vn::VarName{sym}, value) where {sym}
...
+    return VarName{sym}(getoptic(vn) ⨟ optic)
```
devmotion, line 157:
```suggestion
```

src/varname.jl
```diff
@@ -148,6 +149,13 @@ function Accessors.set(obj, vn::VarName{sym}, value) where {sym}
...
+end
```
devmotion, line 158:
```suggestion
```

PR #96, opened 2024-04-10
Ignore `identity` optic when composing

sunxd3:
In `Setfield`, compose a lens with `IdentityLens` will give back the lens, ref
https://github.com/jw3126/Setfield.jl/blob/e31e1e361bd6a251124aa578122e6198b81197b5/src/lens.jl#L148-L152

This behavior can't be reimplemented in `Accessors` because `identity` is a `Base` function.

This PR is for backward compatibility.


torfjelde (approved):
LGTM!

PR #100, opened 2024-09-12
Conversion of VarName to/from string

### LINKED ISSUE #98: `serialisation` / `deserialisation` of `VarName`
Currently, no mechanism exists to serialise `VarName` objects into a string, making it awkward to save DynamicPPL `VarInfo` and MCMCChains objects (see, e.g. [here](https://github.com/TuringLang/Turing.jl/pull/2078)). Thus, it would be helpful to implement helper functions to serialise `VarNames` to string objects and deserialise such strings if needed.

penelopeysm:
Closes #98.


penelopeysm:
@sunxd3 @yebai I'm not entirely sure what interface we want for this, so would like to ask you all for feedback. You can see what it currently does in [the tests I added](https://github.com/TuringLang/AbstractPPL.jl/pull/100/files#diff-76434f1537bbe7bc415b37aabbe159100e04110b2e94bc12783d6137eebd6b18). Specifically:

 - I've chosen to do `varname_from_str()`, but would it be better to instead make a new method `VarName(::AbstractString)`?
 - Not sure what we want to do with `ConcretizedSlice`s. I have half a mind that we should serialise them concretely, i.e. `x = ones(10); @varname(x[:])` should be serialised as `x[1:10]`, as otherwise this loses information.

penelopeysm:
> `Base.Serialization`

I had a look at that earlier, but it skips the 'string' bit and goes straight to 'IO stream'

```julia
serialize(stream::IO, value)
deserialize(stream)
```

meaning that it's not possible to use these methods as part of something larger (e.g. de/serialising a struct that itself contains `VarName`s). Basically it's not composable, or I think, very difficult to use in a composable manner.

Other serialisation libraries like [Serde](https://github.com/bhftbootcamp/Serde.jl) are composable, but we probably don't want to bring in another dependency :(

> internal structure

By this, I'm talking about storing the exact types that VarName uses internally, which would let us correctly differentiate between concretised colons and abstract colons.

```julia
julia> using AbstractPPL; x = ones(15)

julia> dump(@varname(x[:], true))
VarName{:x, Accessors.IndexLens{Tuple{AbstractPPL.ConcretizedSlice{Int64, Base.OneTo{Int64}}}}}
  optic: Accessors.IndexLens{Tuple{AbstractPPL.ConcretizedSlice{Int64, Base.OneTo{Int64}}}}
    indices: Tuple{AbstractPPL.ConcretizedSlice{Int64, Base.OneTo{Int64}}}
      1: AbstractPPL.ConcretizedSlice{Int64, Base.OneTo{Int64}}
        range: Base.OneTo{Int64}
          stop: Int64 15

julia> dump(@varname(x[:]))
VarName{:x, Accessors.IndexLens{Tuple{Colon}}}
  optic: Accessors.IndexLens{Tuple{Colon}}
    indices: Tuple{Colon}
      1: Colon() (function of type Colon)
```

mhauru:
I don't really understand the how concretisation gets used, so no strong opinion. I do agree with @penelopeysm's comment that dumping the entire structure, so one can reconstruct an object that passes a `==` comparison in all cases, sounds like the conceptually correct thing to do.

mhauru:
> an example I can think of is from parameter names in Chain to varname which can be used with VarInfos.

This was my use case that came up.

I think having the same format both machine-readable and human-readable is great, but if that's hard, we can always have one method for a full string representation and another for pretty-printing a summary.

penelopeysm:
Thanks all!

The thing with Chains is that the varname will always be concretised:

```julia
using Turing

@model function demo(y)
    µ ~ Normal(0, 1)
    x = Vector{Real}(undef, 2)
    x[:] ~ MvNormal(fill(µ, 2), 1)
    y ~ Normal(x[end], 1)
end

chain = sample(demo(ones(10)), NUTS(), 100, progress=false)

vn = collect(keys(chain.info.varname_to_symbol))[2]  # x[:][1]
dump(vn)
```

gives

```
AbstractPPL.VarName{:x, ComposedFunction{Accessors.IndexLens{Tuple{Int64}}, Accessors.IndexLens{Tuple{AbstractPPL.ConcretizedSlice{Int64, Base.OneTo{Int64}}}}}}
  optic: (@o _[:][1]) (function of type ComposedFunction{Accessors.IndexLens{Tuple{Int64}}, Accessors.IndexLens{Tuple{AbstractPPL.ConcretizedSlice{Int64, Base.OneTo{Int64}}}}})
    outer: Accessors.IndexLens{Tuple{Int64}}
      indices: Tuple{Int64}
        1: Int64 1
    inner: Accessors.IndexLens{Tuple{AbstractPPL.ConcretizedSlice{Int64, Base.OneTo{Int64}}}}
      indices: Tuple{AbstractPPL.ConcretizedSlice{Int64, Base.OneTo{Int64}}}
        1: AbstractPPL.ConcretizedSlice{Int64, Base.OneTo{Int64}}
          range: Base.OneTo{Int64}
            stop: Int64 2
```

So for this usecase, it seems sensible to say that roundtrip deserialisation need only work for concretised varnames?

torfjelde:
I'm personally very much in favour of storing a "true" representation of `VarName` that can be deserialized to pass something like `==`. It's unclear to me why we want readability in this case?

torfjelde:
> but do we want to use this "string-ify" also for the String cast/conversion like that of Symbol(varname)?

Depends on what the application for this is (I'm somewhat OOL here).

IMO, there are currently two "use-cases" for `string(varname)`:
1. `show`
2. As keys in MCMCChains.jl

(1) is fair, but requires no notion of "deserialization".

(2) is not a good motivation in my opinion. Instead, we should just be using `VarName` directly as keys in `MCMCChains.Chains`. `VarName` implements a `hash`, so AFAIK there's no reason why we wouldn't allow this.

penelopeysm:
Current behaviour is much improved, I think, and works for both unconcretised and concretised slices:

```julia
using AbstractPPL
using Printf

y = ones(10)
vns = [
    @varname(x),
    @varname(x.a),
    @varname(x.a.b),
    @varname(var"x.a"),
    @varname(x[1]),
    @varname(x[1:10]),
    @varname(x[1, 2]),
    @varname(x[:]),
    @varname(y[begin:end]),
    @varname(y[:], false),
    @varname(y[:], true),
]
for vn in vns
    @printf "%10s => %s\n" vn vn_to_string(vn)
end
```

```
         x => (sym = "x", optic = (type = "identity",))
       x.a => (sym = "x", optic = (type = "property", field = "a"))
     x.a.b => (sym = "x", optic = (type = "composed", outer = (type = "property", field = "b"), inner = (type = "property", field = "a")))
       x.a => (sym = "x.a", optic = (type = "identity",))
      x[1] => (sym = "x", optic = (type = "index", indices = "(1,)"))
   x[1:10] => (sym = "x", optic = (type = "index", indices = "(1:10,)"))
   x[1, 2] => (sym = "x", optic = (type = "index", indices = "(1, 2,)"))
      x[:] => (sym = "x", optic = (type = "index", indices = "(:,)"))
   y[1:10] => (sym = "y", optic = (type = "index", indices = "(1:10,)"))
      y[:] => (sym = "y", optic = (type = "index", indices = "(:,)"))
      y[:] => (sym = "y", optic = (type = "index", indices = "(ConcretizedSlice(Base.OneTo(10)),)"))
```

src/varname.jl
```diff
@@ -747,3 +748,62 @@ function vsym(expr::Expr)
...
+
+Convert a string representation of a `VarName` back to a `VarName`.
+
+NOTE: This function should only be used with trusted input, as it uses `eval`
+and `Meta.parse` to parse the string.
+"""
+function vn_from_string(str)
+    new_fields = eval(Meta.parse(str))
+    return VarName{Symbol(new_fields.sym)}(nt_to_optic(new_fields.optic))
+end
```
penelopeysm, line 809: I don't entirely like this, but the only reasonable alternative is to handwrite a parser and manually allow a curated list of expressions to be `eval`'d.

If our main use case is for serialising Chains and other Turing.jl objects to disk, then I think it is reasonable to expect the user to make sure their own input is trusted.

penelopeysm: I'm wary of pulling in a serialisation library – firstly, it's an extra dep for this library, and secondly I'm not sure how it will work with the lenses that we are trying to serialise – do we have to also define serialisation methods for those types (which means either type piracy on our end, or including that functionality in Accessors itself, which is an extra dep for them)? I could be wrong on that point, but that was my superficial understanding from a brief look.

penelopeysm:
```julia
julia> @benchmark map(vn_from_string, strs)
BenchmarkTools.Trial: 1649 samples with 1 evaluation.
 Range (min … max):  2.873 ms … 74.691 ms  ┊ GC (min … max): 0.00% … 95.75%
 Time  (median):     2.920 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.032 ms ±  1.772 ms  ┊ GC (mean ± σ):  1.53% ±  2.80%

 Memory estimate: 242.80 KiB, allocs estimate: 4126.

julia> @benchmark map(vn_from_string2, strs)
BenchmarkTools.Trial: 6158 samples with 1 evaluation.
 Range (min … max):  764.583 μs …  70.964 ms  ┊ GC (min … max): 0.00% … 98.60%
 Time  (median):     777.500 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   810.864 μs ± 900.711 μs  ┊ GC (mean ± σ):  1.77% ±  2.85%

 Memory estimate: 106.89 KiB, allocs estimate: 1614.
```

penelopeysm:
Remaining todos if we're happy with the implementation:

- [ ] Clean up any unused code from old implementation
- [ ] Document
- [ ] Bump Project.toml

torfjelde:
At this point, would it make sense to just move all the varname related stuff to a separate package, e.g. VarNames.jl? This way we could easily use it in MCMCChains for example.

penelopeysm:
> adding StructTypes.jl as a conditional dependency in Accessors.jl

Hmm yes an AccessorsStructTypesExt would make sense! One problem is that we are only really implementing serialisation for a subset of the types in Accessors (i.e. the types we care about). I could open an issue to see whether they would be interested in the code.

> a separate package, e.g. VarNames.jl

AbstractPPL doesn't really have much code that isn't in `varname.jl`, so moving it elsewhere would leave this really empty. I wasn't around for the discussions about why this package was created, though, so happy to leave this decision to you all 😄

penelopeysm:
Actually, coming back to this with a fresh pair of eyes, here's an even faster one which just goes via Dict -> JSON instead of using StructTypes:

```julia
julia> @benchmark map(vn_from_string2, strs)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  68.208 μs …  62.339 ms  ┊ GC (min … max): 0.00% … 99.75%
 Time  (median):     70.583 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   77.770 μs ± 623.365 μs  ┊ GC (mean ± σ):  8.61% ±  1.89%

 Memory estimate: 42.80 KiB, allocs estimate: 677.
```

torfjelde:
> I am probably missing something, why is it difficult now? (I think MCMCChains and AbstractPPL don't depend on each other)

AbstractPPL.jl has a bunch of features that are completely unnecessary for MCMCChains.jl. There are several places where it would be beneficial to use `VarName`s but just seems "too much" to depend on AbstractPPL.jl.

torfjelde:
> Actually, coming back to this with a fresh pair of eyes, here's an even faster one which just goes via Dict -> JSON instead of using StructTypes:

Do we lose some extensibility here though? My "hope" was that if the "serialization interface" is external from AbstractPPL.jl, we'd be able to suggest other packages to also implement it through and extension or something, leading to `VarName` being fully serializable even when more complex indices, etc. are used in a `VarName`.

penelopeysm:
Actually the StructTypes implementation wasn't very extensible either 😅 because it doesn't use StructTypes 'all the way down' – it just uses it for the top layer, and the bottom layers (most importantly, indices) are serialised manually:

https://github.com/TuringLang/AbstractPPL.jl/blob/73a426b04339cf9d7b56c55122e41558295cf855/src/varname.jl#L866-L899

For composability purposes I see a couple of options we can do:

1. Rewrite such that it _does_ use StructTypes all the way down. If we want to avoid type piracy, this would mean we do have to patch everything upstream: for example Accessors types would go in `AccessorsStructTypesExt`, OffsetArrays would go in `OffsetArraysStructTypesExt`, and Base range types would go into `StructTypes` itself.

2. Make some small modifications to the two functions above so that they can be extended to new types by defining a new method. The responsibility for hosting the code for new types would then fall on us: for example, if we wanted to add OffsetArrays compatibility we would add an `AbstractPPL.index_to_dict(::OffsetArrays.IdOffsetRange)` (and the inverse) into an `AbstractPPLOffsetArraysExt`.

Option 1 would be the correct solution if we believe that there is one true serialisation method for those upstream types, which every consumer should work with. I think this is a noble aim! but I feel very hesitant in proposing so many upstream changes :(

As a case study, our ConcretizedSlices (usually) wrap `Base.OneTo`, and it turns out that StructTypes already defines a serialisation method for that:

```julia
julia> using StructTypes, JSON3

julia> s = JSON3.write(Base.OneTo(5))
"[1,2,3,4,5]"
```

Unfortunately, deserialising this results in a loss of information, because there is no information in the string that says that it should be converted to `Base.OneTo{Int}`.

```julia
julia> JSON3.read(s)
5-element JSON3.Array{Int64, Base.CodeUnits{UInt8, String}, Vector{UInt64}}:
 1
 2
 3
 4
 5
```

and although these are semantically the same, it means we can't actually reconstruct a ConcretizedSlice:

```
julia> ConcretizedSlice(Base.Slice(JSON3.read(s)))
ERROR: MethodError: no method matching Base.Slice(::JSON3.Array{Int64, Base.CodeUnits{UInt8, String}, Vector{UInt64}})

Closest candidates are:
  Base.Slice(::Base.Slice)
   @ Base indices.jl:377
  Base.Slice(::T) where T<:AbstractUnitRange
   @ Base indices.jl:375
```

To fix this, we would have to patch StructTypes to serialise `Base.OneTo(5)` as something like

```julia
"{\"stop\":5,\"type\":\"Base.OneTo\"}"
```

so that we could use that `type` information in the deserialisation. And this is the sort of thing I'm hesitant to do – it would be much 'safer' to keep the serialisation function within AbstractPPL, which lets us be more opinionated and customise the exact behaviour to meet _our_ needs.

Let's discuss this on Monday though :)

penelopeysm:
I think this is ready for a final review. I wrote up docs here too which show how you can extend the interface to handle custom types: http://turinglang.org/AbstractPPL.jl/previews/PR100/api/#VarName-serialisation

PR #105, opened 2024-10-06
Update CompatHelper.yml

penelopeysm:


mhauru:
So we used to have two secret SSH keys, one for CompatHelper and one for Documenter. And something has happened to the former (expired?) and thus this PR starts using the latter for CompatHelper as well? Any reason to not add another key CompatHelper? Just a faff, and this gets it done easier? Excuse the volley of questions, just want to make sure I understand what's going on here.

penelopeysm:
> And something has happened to the former (expired?) and thus this PR starts using the latter for CompatHelper as well?

I don't entirely know what happened to make it no longer work, but my suspicion is that GitHub permission-y stuff changed (new changes happen with surprising frequency).

> Any reason to not add another key CompatHelper? Just a faff, and this gets it done easier?

Getting CompatHelper and Documenter to share an SSH key is pretty much the way everyone does it nowadays in Julia (including [the CompatHelper repo](https://github.com/JuliaRegistries/CompatHelper.jl/blob/master/.github/workflows/CompatHelper.yml) itself 😅), most of the repos in TuringLang follow this pattern too.

> Excuse the volley of questions, just want to make sure I understand what's going on here.

No problem :D

mhauru (approved):
Well if that's how CompatHelper itself does it...

PR #109, opened 2024-10-25
Add `fix` and `unfix` functions for parameter fixing and unfixing

sunxd3:


torfjelde (approved):

src/abstractprobprog.jl
```diff
@@ -62,6 +62,46 @@ should hold for generative models `m` and arbitrary `obs`.
 """
 function condition end
 
+"""
+    fix(model, params)
+
+Fix the values of parameters specified in `params` within the probabilistic model `model`. 
+This operation is equivalent to treating the fixed parameters as being drawn from a point mass 
+distribution centered at the values specified in `params`.
```
penelopeysm, line 70: Would it be possible to add a sentence saying that the values of fixed parameters contribute nothing to the log density when the model is evaluated (thus differentiating it from `condition`)?

sunxd3: done

penelopeysm (approved):

PR #110, opened 2024-11-06
Use new GitHub actions

penelopeysm:
We now have some nice actions here that seek to abstract common workflows into the same place: https://github.com/TuringLang/actions

See the repository README for more info.


penelopeysm:
Voila: http://turinglang.org/AbstractPPL.jl/previews/PR110/

sunxd3:
This is great! Sorry I missed it, anything stop us from merging? @penelopeysm

PR #112, opened 2024-12-17
Renaming `decondition` to `uncondition`

### LINKED ISSUE #84: Rename `decondition` to `uncondition`
With `fix` and `unfix` coming in through TuringLang/DynamicPPL.jl#488 , to stay consistent with the naming, we should rename `decondition` to `uncondition`.

sunxd3:
fix #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Therefore, the interface consists of an `AbstractProbabilisticProgram` supertype
 functions
 
   - `condition(::Model, ::Trace) -> ConditionedModel`
-  - `decondition(::ConditionedModel) -> GenerativeModel`
+  - `uncondition(::ConditionedModel) -> GenerativeModel`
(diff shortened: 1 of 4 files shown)


sunxd3:
@penelopeysm this discussion might be before you join, curious for your thoughts on this issue

penelopeysm:
I'm kind of on the fence about this, I get that the naming is more consistent with `unfix` but it doesn't feel like a good enough reason to make a breaking change.

If we go ahead with this, let's at least make sure that we have a plan for percolating this up to DynamicPPL and Turing. Are we going to keep a deprecated implementation of `decondition` around like this?

```julia
Base.@deprecate decondition(model) uncondition(model)
```

sunxd3:
On second thought, I like `decondition` better than `uncondition`. The reason is `de-` means "removal", but "un-" doesn't. Also Penny made a good point here, I am not sure the naming consistency is really worth it for a: (1) arguably worse name (2) breaking release from Turing&DPPL.

@yebai I would suggest just close the issue instead.

PR #121, opened 2025-04-19
Make `prefix` more type-stable

penelopeysm:


penelopeysm:
I thought this would make a difference, but I can't actually see any difference with `@inferred` or `@code_warntype` on Julia 1.11.5. I suppose that the compiler is smart enough to figure out that `getoptic(vn)` returns the second type parameter of `vn`, and thus when it looks at the line `if getoptic(vn) == identity` it can statically determine whether this is true or false.

I don't know if it's risky to rely on this behaviour.

yebai:
It's probably always a good idea to make the compiler's life easier where possible.

sunxd3 (commented):
it looks fine to me and I don't see a downside for this over the original implementation.

Regarding fixing type stability errors, I has couple of thoughts, none of them are too serious:
1. > It appears to me that the compiler is smart enough to figure out that getoptic(vn) returns the second type parameter of vn, and thus when it looks at the line if getoptic(vn) == identity it can statically determine whether this is true or false, and infer the return type appropriately.
    
    is it possible to test the behavior of the compiler in some way? (I am thinking inspecting `IRCode`)
2.  (I don't have a good enough knowledge of the Julia compiler.) `Topic_prefix` is nested, right? Do we have an examples of reasonably deeply nested VarName? So that we can know that the compiler doesn't give up too easily?

PR #123, opened 2025-07-05
Normalise optics when constructing VarName

### LINKED ISSUE #122: nasty edge cases with `prefix` and `unprefix`
```julia
julia> vnp = prefix(@varname(p), @varname(s[1].a))
s[1].a.p

julia> subsumes(@varname(s[1].a), vnp)
false
```

This arises because of the associativity of ComposedFunction -- for example in `s[1].a.p` the lens can be either `ComposedFunction(ComposedFunction(.p, .a), .[1]))` or it could be `ComposedFunction(.p, ComposedFunction(.a, .[1]))` (full functions not shown for simplicity).

There are two ways to handle this. One is to carefully disambiguate all cases with lots of if/else or multiple dispatch methods. The other is to enforce normalisation such that ComposedFunction can only have ComposedFunction as its outer optic (for example) and not as its inner optic. That would require us to either write our own ComposedFunction or upstream it.

penelopeysm:
Closes #122.

I also added a changelog. I'm going to copy paste from it:

- Removed the constructors `VarName(vn, optic)` (this wasn't deprecated, but is IMO quite dangerous as it silently discards the existing optic in `vn`), and `VarName(vn, ::Tuple)` (which was deprecated).
- In the inner constructor of a VarName, its optic is now normalised to ensure that the associativity of ComposedFunction is always the same, and that compositions with identity are removed. This helps to prevent subtle bugs where VarNames with semantically equal optics are not considered equal.


sunxd3 (approved):
lgtm

maybe add a comment somewhere saying that `ComposedOptics` is just a alias of `ComposedFunctions`? (I opted for `ComposedOptics` when migrating to Accessors because I thought it's simpler to understand in context of the migration. But I support use `ComposedFunction` now.)

penelopeysm:
> (I opted for ComposedOptics when migrating to Accessors because I thought it's simpler to understand in context of the migration. But I support use ComposedFunction now.)

That makes sense. I guess it's really Julia that had me confused 😄 , since Accessors writes `const ComposedOptic = ComposedFunction` whenever you dump something it prints as ComposedFunction.

```julia
julia> using Accessors

julia> @o _.a.b
(@o _.a.b)

julia> dump(ans)
(@o _.a.b) (function of type ComposedFunction{PropertyLens{:b}, PropertyLens{:a}})
  outer: PropertyLens{:b} (@o _.b)
  inner: PropertyLens{:a} (@o _.a)
```

I'll add a note!

PR #125, opened 2025-07-05
Move `hasvalue` and `getvalue` from DynamicPPL; implement extra Distributions-based methods

penelopeysm:
## Moving from DynamicPPL

This PR moves `hasvalue` and `getvalue` from DynamicPPL to AbstractPPL. https://github.com/TuringLang/DynamicPPL.jl/blob/92f6eea8660be2142fa4087e5e025f37026bfa45/src/utils.jl#L763-L954

A lot of the helper functions in DynamicPPL are not actually needed because there is existing functionality in here that accomplishes much the same, so I essentially reimplemented both functions.

## Distributions-based methods

This part is new and warrants more explanation. Notice the default behaviour of `hasvalue`:

```julia
julia> d = Dict(@varname(x[1]) => 1.0, @varname(x[2]) => 1.0)
Dict{VarName{:x, Accessors.IndexLens{Tuple{Int64}}}, Float64} with 2 entries:
  x[1] => 1.0
  x[2] => 1.0

julia> hasvalue(d, @varname(x))
false
```

This makes sense, because `d` alone does not give us enough information to reconstruct some arbitrary variable `x`.

However, let's say that we know `x` is to be sampled from a given distribution `dist`. In this case, we do have enough information to determine whether `x` can be reconstructed. This PR therefore also implements the following method:

```julia
julia> using Distributions, LinearAlgebra

julia> hasvalue(d, @varname(x), MvNormal(zeros(2), I))
true

julia> hasvalue(d, @varname(x), MvNormal(zeros(3), I))
false
```

The reason for having this method is to (**properly**) fix issues where values for multivariate distributions are given separately, see e.g., https://github.com/TuringLang/DynamicPPL.jl/issues/712, https://github.com/TuringLang/DynamicPPL.jl/issues/814.

One might argue that we should force users to specify things properly, i.e., if `x ~ MvNormal(zeros(2), I)` then the user should condition on `Dict(@varname(x) => [1.0, 2.0])` rather than the dict above. This is fine, and in an ideal world I would do that. **However, there is one specific case where this isn't enough,** namely in DynamicPPL's `predict(model, chain)` or `returned(model, chain)`. Unfortunately, `chain` is a lossy storage format, because vector variables like `x` are split up into `x[1]` and `x[2]` and it's not possible to recover the original shape of `x`.

Up until this PR, this has been handled in DynamicPPL using the `setval_and_resample!` and `nested_setindex_maybe` methods which perform some direct manipulation of VarInfos. It is also dangerous in that if only _part_ of the variable `x` is given, it marks the entire variable `x` as to be not-resampled, which IMO can lead to silent bugs: https://github.com/TuringLang/DynamicPPL.jl/blob/92f6eea8660be2142fa4087e5e025f37026bfa45/src/varinfo.jl#L2177-L2181

With this `hasvalue` and `getvalue` method we can deal with this case in a more principled fashion. If we can find all the indices needed to reconstruct the value of `x`, then we can set that value properly; if we can't, then we _don't_ set any of them.

## TODO

- [ ] Can we simplify the API? Note that `hasvalue` and `getvalue` have extremely similar logic, do we really need to have two functions with almost the same implementation? The only concern I have is type stability, i.e. `getvalue` is inherently type-unstable, and maybe guarding calls to `getvalue` behind a call to `hasvalue` avoids leaking type instability into the caller function. However, I think this is reliant on the compiler being able to infer the return value of `hasvalue` through e.g. constant propagation?!
- [ ] `hasvalue` for other distributions
- [ ] `getvalue` for distributions
- [ ] Appropriate tests
- [ ] API documentation for the distributions bits
- [ ] Changelog

Note that `InitContext` https://github.com/TuringLang/DynamicPPL.jl/pull/967 is blocked by this.

In turn this PR requires #123 to be merged first as without the normalisation, `haskey(dict, vn)` can run into issues if `vn`'s optic is not normalised.


src/varname.jl
```diff
@@ -963,69 +963,147 @@ string_to_varname(str::AbstractString) = dict_to_varname(JSON.parse(str))
...
-function _strip_identity(o::Base.ComposedFunction{Outer,typeof(identity)}) where {Outer}
-    return _strip_identity(o.outer)
-end
-function _strip_identity(o::Base.ComposedFunction{typeof(identity),Inner}) where {Inner}
-    return _strip_identity(o.inner)
-end
-_strip_identity(o::Base.ComposedFunction) = o
-_strip_identity(o::Accessors.PropertyLens) = o
-_strip_identity(o::Accessors.IndexLens) = o
-_strip_identity(o::typeof(identity)) = o
```
penelopeysm, line 980: `normalise` strips identities now so this function isn't needed any more

src/varname.jl
```diff
@@ -963,69 +963,147 @@ string_to_varname(str::AbstractString) = dict_to_varname(JSON.parse(str))
...
 ```
 """
-_inner(o::Base.ComposedFunction{Outer,Inner}) where {Outer,Inner} = o.inner
-_inner(o::Accessors.PropertyLens) = o
-_inner(o::Accessors.IndexLens) = o
-_inner(o::typeof(identity)) = o
+_head(o::ComposedFunction{Outer,Inner}) where {Outer,Inner} = o.inner
+_head(o::Accessors.PropertyLens) = o
+_head(o::Accessors.IndexLens) = o
+_head(::typeof(identity)) = identity
```
penelopeysm, line 998: `_head`, `_tail`, `_init`, and `_last` take their names from the equivalent Haskell functions on linked lists:

```haskell
λ> head [1,2,3]
1
λ> tail [1,2,3]
[2,3]
λ> init [1,2,3]
[1,2]
λ> last [1,2,3]
3

-- empty list is turned into identity in our case
λ> head [1]
1
λ> tail [1]
[]
λ> init [1]
[]
λ> last [1]
1
```

ext/AbstractPPLDistributionsExt.jl
```diff
@@ -0,0 +1,276 @@
...
+    is_up = dist.uplo == 'U'
+    cartesian_indices = filter(CartesianIndices(size(dist))) do cartesian_index
+        i, j = cartesian_index.I
+        is_up ? i <= j : i >= j
+    end
+    # there is an additional layer as we need to access `.L` or `.U` before we
+    # can index into it
+    field_lens = is_up ? (Accessors.@o _.U) : (Accessors.@o _.L)
+    return map(idx -> Accessors.IndexLens(idx.I) ∘ field_lens, cartesian_indices)
+end
```
mhauru, line 56: I'm confused by the scenario in which we need this. LKJCholesky returns objects of type `Cholesky`, which are not `AbstractArray`s and can't be indexed. Would we ever have a situation where we would have something like
```
getvalue(Dict(@varname(x[1,1]) => 1.0, @varname(x[1,2]) => 0.0, @varname(x[2,2]) => 1.0), @varname(x), LKJCholesky(2, 0.5))
```
where it should return `true`?

PS. After reading the tests I know realise you need things like `@varname(x.U[1,1])`, but I still wonder if this would ever come up.

ext/AbstractPPLDistributionsExt.jl
```diff
@@ -0,0 +1,276 @@
...
+    expected_vns = map(o -> VarName{sym}(o ∘ original_optic), optics)
+    if all(sub_vn -> AbstractPPL.hasvalue(vals, sub_vn), expected_vns)
+        return true
+    else
+        if error_on_incomplete &&
+            any(sub_vn -> AbstractPPL.hasvalue(vals, sub_vn), expected_vns)
+            error("hasvalue: only partial values for `$vn` found in the values provided")
+        end
+        return false
+    end
```
mhauru, line 183: What happens, and what should happen, with the following?
```julia
hasvalue(
    Dict(@varname(x) => [1.0], @varname(x[2]) => 2.0),
    @varname(x),
    MvNormal(zeros(2), I),
)
```

I think this returns `true` on line 163 even though the shape of `@varname(x) => [1.0]` doesn't match the distribution. But even if it didn't, I think it would it also return true because `@varname(x[1])` is found in `@varname(x) => [1.0]` and `@varname(x[2])` is found in `@varname(x[2]) => 2.0`. That doesn't quite feel right though.

test/hasvalue.jl
```diff
@@ -0,0 +1,209 @@
+@testset "base getvalue + hasvalue" begin
+    @testset "basic NamedTuple" begin
+        nt = (a=[1], b=2, c=(x=3, y=[4], z=(; p=[(; q=5)])), d=[1.0 0.5; 0.5 1.0])
```
mhauru, line 3:
```suggestion
        nt = (
            a=[1],
            b=2,
            c=(
                x=3,
                y=[4],
                z=(;p=[(; q=5)])
            ),
            d=[1.0 0.5; 0.5 1.0],
        )
```
This whole NamedTuple just made me laugh. I also found it really hard to read. I wrote that suggestion, but I guess autoformatter is just gonna brutalise it back into unreadability, so maybe this is futile.

test/hasvalue.jl
```diff
@@ -0,0 +1,209 @@
...
+        @test !hasvalue(d, @varname(c.y.a))
+        @test !hasvalue(d, @varname(c.zzzz))
+        @test !hasvalue(d, @varname(d[1, 3]))
+    end
+
+    @testset "Dict with non-identity varname keys" begin
+        d = Dict(
+            @varname(a[1]) => [1.0, 2.0],
+            @varname(b.x) => [3.0],
+            @varname(c[2]) => (a=4.0, b=5.0),
```
mhauru, line 105: What about having this be a dict within a dict?

test/hasvalue.jl
```diff
@@ -0,0 +1,209 @@
...
+        @test hasvalue(d, @varname(x), Normal())
+        @test getvalue(d, @varname(x), Normal()) == 1.0
+        @test hasvalue(d, @varname(y[1][1]), Normal())
+        @test getvalue(d, @varname(y[1][1]), Normal()) == 2.0
+    end
+
+    @testset "multivariate + matrix" begin
+        d = Dict(@varname(x[1]) => 1.0, @varname(x[2]) => 2.0)
+        @test hasvalue(d, @varname(x), MvNormal(zeros(1), I))
+        @test getvalue(d, @varname(x), MvNormal(zeros(1), I)) == [1.0]
```
mhauru, line 168: It's not wrong, but it feels funny.

mhauru (commented):
Happy with the code except for some minor localised comments, but confused about whether we really need to deal with LKJCholesky, and thus whether we could restrict ourselves to just caring about dimensions rather than more complicated forms of distribution outputs.

ext/AbstractPPLDistributionsExt.jl
penelopeysm: I guess this ambiguity could come up if the user manually mis-specified without the `.L` or `.U`. However, assuming that the varnames come from MCMCChains, then the varnames will have been constructed correctly (via `varname_and_value_leaves`... which ALSO special-cases Cholesky https://github.com/TuringLang/DynamicPPL.jl/blob/ce7c8b1ae48624e12ebf3064b6099e3dfca8c985/src/utils.jl#L1258-L1265)

HISTORY.md
```diff
@@ -1,3 +1,16 @@
+## 0.12.1
+
+Minimum compatibility has been bumped to Julia 1.10.
```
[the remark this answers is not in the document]
penelopeysm: Happy to do a minor bump due to the large amount of code additions.

test/hasvalue.jl
penelopeysm: Hold on, I can't get it to work because you can't use a PropertyLens to index into a dictionary.

```julia
julia> d = Dict(@varname(a) => Dict(@varname(b) => 1))
Dict{VarName{:a, typeof(identity)}, Dict{VarName{:b, typeof(identity)}, Int64}} with 1 entry:
  a => Dict(b=>1)

julia> d.a
ERROR: type Dict has no field a
Stacktrace:
 [1] getproperty(x::Dict{VarName{:a, typeof(identity)}, Dict{VarName{:b, typeof(identity)}, Int64}}, f::Symbol)
   @ Base ./Base.jl:49
 [2] top-level scope
   @ REPL[4]:1

julia> d = Dict(:a => Dict(:b => 1))
Dict{Symbol, Dict{Symbol, Int64}} with 1 entry:
  :a => Dict(:b=>1)

julia> d.a
ERROR: type Dict has no field a
Stacktrace:
 [1] getproperty(x::Dict{Symbol, Dict{Symbol, Int64}}, f::Symbol)
   @ Base ./Base.jl:49
 [2] top-level scope
   @ REPL[6]:1
```

So nope, `hasvalue` can't handle dicts in dicts (unless we recursively call `hasvalue`).

src/hasvalue.jl
```diff
@@ -0,0 +1,262 @@
...
+ 1.0
+
+julia> getvalue(vals, @varname(x[1])) # different from `getindex`
+1.0
+
+julia> getvalue(vals, @varname(x[2]))
+ERROR: getvalue: x[2] was not found in the values provided
+[...]
+```
+
```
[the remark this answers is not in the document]
penelopeysm: Sure. Would you rather have it in the docstring, or the test suite?

ext/AbstractPPLDistributionsExt.jl
penelopeysm: (cf. comment below in main thread)

test/hasvalue.jl
penelopeysm: (cf. comment below in main thread)

HISTORY.md
```diff
@@ -1,3 +1,16 @@
...
+
+Added the new functions `hasvalue(container::T, ::VarName[, ::Distribution])` and `getvalue(container::T, ::VarName[, ::Distribution])`, where `T` is either `NamedTuple` or `AbstractDict{<:VarName}`.
+
+These functions check whether a given `VarName` has a value in the given `NamedTuple` or `AbstractDict`, and return the value if it exists.
+
+The optional `Distribution` argument allows one to reconstruct a full value from its component indices.
+For example, if `container` has `x[1]` and `x[2]`, then `hasvalue(container, @varname(x), dist)` will return true if `size(dist) == (2,)` (for example, `MvNormal(zeros(2), I)`).
```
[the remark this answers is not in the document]
penelopeysm: We were so close, it's just Cholesky that breaks it.

test/hasvalue.jl
mhauru: That's fine, good to know though.

src/hasvalue.jl
```diff
@@ -0,0 +1,262 @@
...
+2.0
+```
+"""
+function getvalue(vals::NamedTuple, vn::VarName{sym}) where {sym}
+    optic = getoptic(vn)
+    if haskey(vals, sym) && canview(optic, getproperty(vals, sym))
+        return optic(vals[sym])
+    else
+        error("getvalue: $(vn) was not found in the values provided")
+    end
```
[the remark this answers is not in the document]
penelopeysm: Error message changed, I think I do prefer a bit not mentioning the function name (it's in the stacktrace anyway). 

Re. `canview`: I think this ties into one of the things I wasn't fully sure about (on the main comment).

The idea is that you could use `getvalue` without having checked `hasvalue`, so in principle you do need to check bounds in both. That is very annoying not only because of the double bounds check but also because of the severe code duplication (notice that this PR could have been half the lines if we combined the two functions into one).

I don't fully know how to solve this though. The obvious solution would be for `getvalue` to return a sentinel value if it's not found. However I worry that that would introduce type instability.

HISTORY.md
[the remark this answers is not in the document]
mhauru: Somehow dealing with a sample also just feel dirty and wrong. The
```julia
julia> size(LKJCholesky(2, 1.0))
(2, 2)
```
thing just feels wrong, or maybe rather we need `Distributions.sample_type`, but I guess that isn't ours to fix. Sigh, so be it then.

mhauru (approved):
Thanks!

PR #154, opened 2026-04-04
Micro-optimise append_optic; fix serialisation for JSON@1.5.0

penelopeysm:
shortcircuit the recursion in a simple case


penelopeysm:
honestly this is not even a micro-optimisation, it's more like a nano-optimisation

~~the CI failure is bizarre though, i can't reproduce it offline~~ Reproduced, the culprit is upgrading JSON.jl from 1.4.0 to 1.5.0

PR #155, opened 2026-04-14
Add evaluator interface with AD backend extensions

yebai:
## Summary                                                                                                                                                   
                                                                  
  Adds a prepared evaluator API to AbstractPPL that decouples log-density                                                                                      
  evaluation from automatic differentiation, following the interface sketched
  in `docs/src/interface.md`. 

  Autograd backends are optional. For example, removing the Enzyme package extension would simply fall back to DifferentiationInterface for `AutoEnzyme`. The broader aim is to rely only on public APIs of autodiff backends, while keeping gradient definitions independent of any particular backend implementation.                                                                                                                  

  - This PR implements this design [note](https://github.com/TuringLang/design-docs/blob/main/designs/0001-named-evaluators.md). 
  - Fix https://github.com/TuringLang/DynamicPPL.jl/issues/880


TODOs
- [X] make sure the evaluator interface is sufficient for DynamicPPL
- [X] make sure the evaluator interface is sufficient for AdvancedVI
- [X] make sure the evaluator interface is sufficient for Bijectors

```julia
 using AbstractPPL, ADTypes

 # 1. Define a problem type and its structural evaluator
 struct MyProblem end
 struct MyEvaluator end

 AbstractPPL.prepare(::MyProblem, ::NamedTuple) = MyEvaluator()
 (::MyEvaluator)(nt::NamedTuple) = nt.x^2 + sum(nt.y .^ 2)

 # 2. Prepare an AD-aware evaluator (works with any supported backend)
 using ForwardDiff
 prepared = prepare(AutoForwardDiff(), MyProblem(), (x=0.0, y=[0.0, 0.0]))

 # 3. Evaluate and differentiate — gradients keep the named structure
 val, grad = value_and_gradient(prepared, (x=3.0, y=[1.0, 2.0]))
 # val    = 14.0
 # grad.x = 6.0
 # grad.y = [2.0, 4.0]

 # 4. Verify gradients against finite differences (in tests)
using FiniteDifferences
 test_autograd(prepared, (x=3.0, y=[1.0, 2.0]))
```


gdalle:
@wsmoses you may want to take a look at this Enzyme ext, especially since the test doesn't fully cover it yet?

ext/AbstractPPLEnzymeExt.jl
```diff
@@ -0,0 +1,52 @@
...
+    length(x) == p.dim ||
+        throw(DimensionMismatch("expected vector of length $(p.dim), got $(length(x))"))
+    return p.f_vec(x)
+end
+
+function AbstractPPL.prepare(::AutoEnzyme, problem, prototype::NamedTuple)
+    evaluator = AbstractPPL.prepare(problem, prototype)
+    x0 = AbstractPPL.flatten_to_vec(prototype)
+    f_vec = let evaluator = evaluator, prototype = prototype
+        x -> evaluator(AbstractPPL.unflatten_from_vec(prototype, x))
```
wsmoses: don't make a closure/capture the evaluator here, just call evaluator directly in autodiff with all of the arguments. Also don't add this extra shim to flatten/unflatten

ext/AbstractPPLDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,46 @@
...
+        x -> evaluator(AbstractPPL.unflatten_from_vec(prototype, x))
+    end
+    prep = DI.prepare_gradient(f_vec, adtype, x0)
+    return DIPrepared(evaluator, f_vec, adtype, prep, prototype, length(x0))
+end
+
+@inline function AbstractPPL.value_and_gradient(p::DIPrepared, values::NamedTuple)
+    x = AbstractPPL.flatten_to_vec(values)
+    val, dx = DI.value_and_gradient(p.f_vec, p.prep, p.backend, x)
+    grad_nt = AbstractPPL.unflatten_from_vec(p.prototype, dx)
```
gdalle: Why flatten and unflatten when you could keep the structured representation?

ext/AbstractPPLEnzymeExt.jl
```diff
@@ -0,0 +1,52 @@
+module AbstractPPLEnzymeExt
+
+using AbstractPPL: AbstractPPL, DerivativeOrder
+using ADTypes: AutoEnzyme
+using Enzyme: Enzyme
+
+struct EnzymePrepared{E,F,T<:Real,P}
+    evaluator::E
+    f_vec::F
+    gradient_buffer::Vector{T}
```
gdalle: Why a `Vector` instead of whatever array type is appropriate (e.g. on GPU)?

ext/AbstractPPLDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,46 @@
...
+struct DIPrepared{E,F,B,C,P}
+    evaluator::E
+    f_vec::F
+    backend::B
+    prep::C
+    prototype::P
+    dim::Int
+end
+
+AbstractPPL.capabilities(::Type{<:DIPrepared}) = DerivativeOrder{1}()
```
gdalle: Some DI backends can go up to order 2

ext/AbstractPPLEnzymeExt.jl
```diff
@@ -0,0 +1,52 @@
...
+    return p.f_vec(x)
+end
+
+function AbstractPPL.prepare(::AutoEnzyme, problem, prototype::NamedTuple)
+    evaluator = AbstractPPL.prepare(problem, prototype)
+    x0 = AbstractPPL.flatten_to_vec(prototype)
+    f_vec = let evaluator = evaluator, prototype = prototype
+        x -> evaluator(AbstractPPL.unflatten_from_vec(prototype, x))
+    end
+    grad_buf = zeros(eltype(x0), length(x0))
```
gdalle:
```suggestion
    grad_buf = zero(x0)
```

ext/AbstractPPLEnzymeExt.jl
```diff
@@ -0,0 +1,52 @@
...
+    return EnzymePrepared(evaluator, f_vec, grad_buf, prototype, length(x0))
+end
+
+@inline function AbstractPPL.value_and_gradient(p::EnzymePrepared, values::NamedTuple)
+    x = AbstractPPL.flatten_to_vec(values)
+    fill!(p.gradient_buffer, 0.0)
```
gdalle:
```suggestion
    fill!(p.gradient_buffer, zero(eltype(x)))
```

ext/AbstractPPLEnzymeExt.jl
```diff
@@ -0,0 +1,52 @@
...
+    result = Enzyme.autodiff(
+        Enzyme.set_runtime_activity(Enzyme.ReverseWithPrimal),
+        Enzyme.Const(p.f_vec),
```
gdalle: Since `f_vec` is a closure, I'm not sure what the proper annotation is here

ext/AbstractPPLFiniteDifferencesExt.jl
```diff
@@ -0,0 +1,43 @@
+module AbstractPPLFiniteDifferencesExt
```
gdalle: This extension contains exactly the same code as the DI extension. What is the point of duplication?

ext/AbstractPPLForwardDiffExt.jl
```diff
@@ -0,0 +1,50 @@
+module AbstractPPLForwardDiffExt
```
gdalle: This is exactly what DI does too. What is the point of duplication?

src/evaluator.jl
```diff
@@ -0,0 +1,121 @@
...
+_scalar_count(x::AbstractArray{<:Real}) = length(x)
+function _scalar_count(nt::NamedTuple)
+    n = 0
+    for v in values(nt)
+        n += _scalar_count(v)
+    end
+    return n
+end
+
+_vec_eltype(x::Real) = typeof(x)
```
gdalle: `eltype` works on numbers too

src/evaluator.jl
```diff
@@ -0,0 +1,121 @@
...
+    return offset + n
+end
+function _flatten!(vec::AbstractVector, nt::NamedTuple, offset::Int)
+    for v in values(nt)
+        offset = _flatten!(vec, v, offset)
+    end
+    return offset
+end
+
+function flatten_to_vec(nt::NamedTuple)
```
gdalle: There are probably packages for flattening stuff more cleanly

src/evaluator.jl
```diff
@@ -0,0 +1,121 @@
...
+    n = _scalar_count(nt)
+    vec = Vector{_vec_eltype(nt)}(undef, n)
```
gdalle: Why use a `Vector` to flatten e.g. a named tuple of GPU arrays?

src/evaluator.jl
```diff
@@ -0,0 +1,121 @@
...
+    _flatten!(vec, nt, 1)
+    return vec
+end
+
+function _unflatten(::Real, vec::AbstractVector, offset::Int)
+    return vec[offset], offset + 1
+end
+function _unflatten(proto::AbstractArray{<:Real}, vec::AbstractVector, offset::Int)
+    n = length(proto)
+    result = reshape(@view(vec[offset:(offset + n - 1)]), size(proto))
```
gdalle: The unflattening doesn't have the same type as the initial input

test/ext/differentiation_interface/differentiation_interface.jl
```diff
@@ -0,0 +1,50 @@
...
+function AbstractPPL.prepare(::QuadraticProblem, prototype::NamedTuple)
+    return QuadraticPrepared()
+end
+
+function (::QuadraticPrepared)(values::NamedTuple{(:x, :y)})
+    return values.x^2 + sum(vi -> vi^2, values.y)
+end
+
+# FiniteDifferences has no native AbstractPPL extension, so AbstractPPLDifferentiationInterfaceExt
+# is the only applicable dispatch path for this backend.
```
gdalle: That is obviously not true, you (or Claude?) literally added that extension above

test/ext/differentiation_interface/differentiation_interface.jl
```diff
@@ -0,0 +1,50 @@
...
+const fdm = FiniteDifferences.central_fdm(5, 1)
+const adtype = ADTypes.AutoFiniteDifferences(; fdm)
+
+@testset "AbstractPPLDifferentiationInterfaceExt" begin
```
gdalle: The test code is duplicated across all backends, which is a bit of a shame

gdalle (commented):
This 1000-LOC PR introduces a lot of duplication across backends, for reasons that I still don't fully understand. Again, if DI is insufficient in any way for Turing's purposes, feel free to point it out here and I'll work on it: https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/992

gdalle (changes requested):
Forgot to pick the correct review option

wsmoses:
@yebai a high level question: why are you doing this?

the APIs here look remarkably similar to differentiation interface, but carrying extra maintenance burden, and also being less optimal than the current DI wrappers?

is there something about the way you’re calling things here that DI doesn’t support (or an extra optimization knowing information about the problem context)?

yebai:
@wsmoses, thank you for your engagement. If you are interested in contributing to the TuringLang codebase, the preferred approach is to draft a feature proposal and submit it to a team member for review. The TuringLang team will then indicate whether the proposal is accepted or rejected. Upon acceptance, you may proceed with implementing the proposed feature.

wsmoses:
Got it @yebai can you please clarify your contribution policy then in the readme and other relevant documentation?

To my mind this is a reasonably normal question for reviewing this PR specifically (for which I was explicitly tagged to review how you modify the usage of Enzyme, which I have spent years alongside @penelopeysm @mhauru @Red-Portal @devmotion and others to get to work nicely with turing, as part of a grant which we are both recipients on to ensure Enzyme and Turing work together: https://www.turing.ac.uk/research/research-projects/development-turingjl-probabilistic-programming-language )? Such a review would also have mitigated the type piracy on Enzyme on Advanced I you pushed to main without review (and subsequently force pushed a removal of).

As such I would have been expected to be the very turinglang team reviewer for this PR which you are contributing as a feature request?

However it's not my decision to determine Turings contribution policy.

Getting clarification on this policy would have certainly helpful to me in this case.

cc @rayegun @kshyatt @goerz @samuelsonic @lkdvos @DilumAluthge and @ChrisRackauckas for visibility since I think your contributor policy has come up in other threads and I think would all benefit from clarity on the issue, rather than having to copy it through different threads

ext/AbstractPPLDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,52 @@
...
+    end
+end
+
+# Catch-all for backends without a native AbstractPPL extension; native
+# extensions take precedence via more-specific positional types.
+# NamedTuple inputs are not handled here; native extensions cover that path.
+function AbstractPPL.prepare(
+    adtype::ADTypes.AbstractADType,
+    problem,
+    x::AbstractVector{<:AbstractFloat};
```
penelopeysm: Is there a reason why `AbstractFloat` rather than `Real`? A looser type bound of Real would allow for e.g. higher-order or sparse AD without affecting the common use case of Float64

ext/AbstractPPLDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,52 @@
...
+    check_dims::Bool=true,
+    mode::Symbol=:gradient,
```
penelopeysm: I think this argument shouldn't be a Symbol (notice that if you pass `mode=:NotARealMode`, it will do the Jacobian which is probably not what was intended). Either a custom type would be better

```julia
abstract type Mode end
struct Gradient <: Mode end
struct Jacobian <: Mode end

struct Foo{F<:Mode}
   ...
end
```

ext/AbstractPPLDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,52 @@
...
+    raw = AbstractPPL.prepare(problem, x)
+    length(x) == 0 && return AbstractPPL.ADProblems.VectorEvaluator{check_dims}(raw, 0)
+    evaluator = AbstractPPL.ADProblems.VectorEvaluator{check_dims}(raw, length(x))
+    if mode === :gradient
+        prep = DI.prepare_gradient(evaluator, adtype, x)
+        return DIPrepared{:gradient}(evaluator, adtype, prep)
+    else
+        prep = DI.prepare_jacobian(evaluator, adtype, x)
+        return DIPrepared{:jacobian}(evaluator, adtype, prep)
+    end
```
penelopeysm, line 58: or if you insist on it being a Symbol, better to be more defensive here and use `elseif mode === :jacobian`. But types are better imo

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
+module ADProblems
```
penelopeysm, line 1: I still think this should be a separate package since, as per https://github.com/chalk-lab/Mooncake.jl/pull/1153#issuecomment-4291924982, the stated aim is to be more general than probabilistic programming. It wouldn't really make sense to put it in a package that's called PPL. It is also less discoverable when it's deep inside another package.
I would really suggest spinning it out now otherwise later on it's another breaking change and more faff to remove it

src/AbstractPPL.jl
```diff
@@ -10,6 +10,11 @@ export AbstractModelTrace
 include("abstractmodeltrace.jl")
 include("abstractprobprog.jl")
 include("evaluate.jl")
+include("ADProblems.jl")
+using .ADProblems: prepare, value_and_gradient, value_and_jacobian, test_autograd
+export prepare, value_and_gradient, value_and_jacobian, test_autograd
```
penelopeysm: Is it mandatory to re-export at the top level? Would it be sufficient to export from the submodule and call it a day?

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+
+using ADTypes: ADTypes
+
+export prepare, value_and_gradient, value_and_jacobian, test_autograd
+
+"""
+    AbstractPrepared{Mode}
+
+Internal abstract supertype for all AD-prepared evaluators produced by AbstractPPL's
+extension backends. `Mode` is `:gradient` or `:jacobian`.
```
penelopeysm: As above, I think a Symbol is too loose

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+on the vector-input path.
+"""
+function prepare end
+
+function _check_mode(mode::Symbol)
+    mode === :gradient ||
+        mode === :jacobian ||
+        throw(ArgumentError("`mode` must be `:gradient` or `:jacobian`, got `:$(mode)`."))
+    return nothing
+end
```
penelopeysm: If you don't use a Symbol, then you don't need these functions

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+"""
+    prepare(problem, values::NamedTuple)
+    prepare(problem, x::AbstractVector{<:AbstractFloat})
+    prepare(adtype, problem, values_or_vector; check_dims::Bool=true, mode::Symbol=:gradient)
+
+Prepare a callable evaluator for `problem`.
+
+Use the two-argument form with a `NamedTuple` when the evaluator works with
+named inputs, or with a vector of floating-point numbers when it works with
+vector inputs. Automatic-differentiation backends extend this interface with
```
penelopeysm:
```suggestion
vector inputs. Automatic differentiation backends extend this interface with
```

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+
+"""
+    prepare(problem, values::NamedTuple)
+    prepare(problem, x::AbstractVector{<:AbstractFloat})
+    prepare(adtype, problem, values_or_vector; check_dims::Bool=true, mode::Symbol=:gradient)
+
+Prepare a callable evaluator for `problem`.
+
+Use the two-argument form with a `NamedTuple` when the evaluator works with
+named inputs, or with a vector of floating-point numbers when it works with
```
penelopeysm: Real maybe?

ext/AbstractPPLLogDensityProblemsExt.jl
```diff
@@ -0,0 +1,43 @@
+module AbstractPPLLogDensityProblemsExt
+
+using AbstractPPL: AbstractPPL
+using AbstractPPL.ADProblems: AbstractPrepared, VectorEvaluator, NamedTupleEvaluator
+using LogDensityProblems: LogDensityProblems
+
+LogDensityProblems.logdensity(p::AbstractPrepared, x) = p(x)
+LogDensityProblems.logdensity(e::VectorEvaluator, x) = e(x)
+LogDensityProblems.logdensity(e::NamedTupleEvaluator, x::NamedTuple) = e(x)
```
penelopeysm: Regarding this overload for NamedTuple: I get the sense that LogDensityProblems isn't really intended to work with NT anyway...? If you want to take NT input, I feel it might be more idiomatic to make your own LogDensityProblems-like interface that works with NT.

ext/AbstractPPLLogDensityProblemsExt.jl
```diff
@@ -0,0 +1,43 @@
+module AbstractPPLLogDensityProblemsExt
```
penelopeysm, line 1: LogDensityProblems is a tiny dep, so I think this could just go in src instead of being an extension

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+
+The `Trivial` type parameter is `true` iff `dim == 0`.
+"""
+struct VectorEvaluator{Checked,Trivial,F}
+    f::F
+    dim::Int
+    function VectorEvaluator{Checked}(f::F, dim::Int) where {Checked,F}
+        Checked isa Bool || throw(ArgumentError("`Checked` must be a Bool."))
+        dim >= 0 || throw(ArgumentError("`dim` must be non-negative, got $dim."))
+        return new{Checked,dim == 0,F}(f, dim)
```
penelopeysm: This is not type stable, because `dim==0` is at the value level. That might or might not be a problem. Do you know if there is any issue with this?

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+Internal evaluator shape for scalar functions of a floating-point vector input.
+Used by AbstractPPL's AD extensions; this is not part of the public API.
+
+`Checked` controls whether the call method validates the input length. The default
+(`true`) is the safe shape exposed to users via `prepared(x)`. AD extensions may
+construct `VectorEvaluator{false}` for the inner callable handed to AD libraries,
+where the input length is already guaranteed and the runtime check would otherwise
+remain in the dual/shadow hot path.
+
+The `Trivial` type parameter is `true` iff `dim == 0`.
```
penelopeysm: This type parameter isn't mentioned in the docstring, so this might be a bit confusing.

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+"""
+    VectorEvaluator{Checked}(f, dim)
+    VectorEvaluator(f, dim)  # equivalent to `VectorEvaluator{true}(f, dim)`
+
+Internal evaluator shape for scalar functions of a floating-point vector input.
+Used by AbstractPPL's AD extensions; this is not part of the public API.
+
+`Checked` controls whether the call method validates the input length. The default
+(`true`) is the safe shape exposed to users via `prepared(x)`. AD extensions may
+construct `VectorEvaluator{false}` for the inner callable handed to AD libraries,
```
penelopeysm, line 66: I don't really understand what "safe shape" means here. Can this be rephrased? Just something like "By default `prepared` returns an evaluator with `Checked=true` as that avoids making assumptions about the user's inputs".
Also, in general I think `Checked` is a bit confusing. If `Checked=true`, it can easily be read as "the length has **already** been checked, and we don't need to do it again". In fact this is how I read it the first time.

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+
+Errors if `prepared` was built with `mode=:jacobian`; only gradient-mode
+evaluators are supported.
+
+Requires loading FiniteDifferences so the extension-backed implementation is available.
+"""
+function test_autograd(prepared, x; atol=1e-5, rtol=1e-5)
+    _assert_gradient_capability(prepared)
+    throw(
+        ArgumentError(
```
penelopeysm: I think it's not really an `ArgumentError` if the backend isn't loaded (there's nothing wrong with the arguments per se). I'm not sure what the right error is, but maybe just `error(...)` is more honest.

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+            "This evaluator does not support jacobians. Re-prepare it with `mode=:jacobian`.",
+        ),
+    )
+end
+
+"""
+    test_autograd(prepared, x; atol=1e-5, rtol=1e-5)
+
+Compare `value_and_gradient(prepared, x)` against a finite-difference reference.
```
penelopeysm: In general it can be necessary to compare against other AD backends (not FiniteDifferences) because of issues like https://github.com/JuliaDiff/FiniteDifferences.jl/issues/52 so I think it would be useful to generalise this to take any generic adtype. In fact one might also just want to pass a pre-computed answer (for example if you are testing many backends and want to only calculate the correct answer once). Maybe https://github.com/TuringLang/DynamicPPL.jl/blob/4201c6ce13c9d5a7d1756ab2a4f9c2564781587e/src/test_utils/ad.jl#L260-L276 is useful for ideas?

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+function value_and_jacobian end
+
+function value_and_jacobian(prepared, x::AbstractVector{<:AbstractFloat})
+    throw(
+        ArgumentError(
+            "This evaluator does not support jacobians. Re-prepare it with `mode=:jacobian`.",
+        ),
+    )
+end
+
```
penelopeysm: I think it would be better, as far as possible, to not define empty methods that only throw an error. The main reasons for this are:

1. It leads to extra boilerplate code that is hard to keep in one's head and to maintain.
2. It can cause method ambiguities quite easily. Consider if someone defines `value_and_jacobian(::MyPrepared, x::AbstractVector{<:Real})`, which is I think quite a reasonable thing to expect to do. That's a method ambiguity. Sometimes method ambiguities indicate a genuine ambiguity that needs to be resolved with a more specific method, but in this case it would force the user to resolve an ambiguity with a trivial method, which is less ideal.

A way around this if you want the error messages is to leave the method undefined, and add an error hint for MethodErrors e.g. https://github.com/TuringLang/DynamicPPL.jl/blob/4201c6ce13c9d5a7d1756ab2a4f9c2564781587e/src/DynamicPPL.jl#L288-L291 or https://docs.julialang.org/en/v1/base/base/#Base.Experimental.register_error_hint

Alternatively, one could at least broaden the type signature to `AbstractVector` as that will minimise the chances of method ambiguities.

docs/src/pplapi.md
penelopeysm, line 1: For such a large PR, I think it would be very beneficial to have docs explaining the motivation behind this, the interface, and some examples of how to use it. (That would also have been very helpful for PR review.) Personally I think this is the biggest thing missing from this PR. There is a lot of code but it should be accompanied by explanation IMO

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+    _assert_gradient_capability(prepared)
+    throw(
+        ArgumentError(
+            "`test_autograd` requires loading FiniteDifferences to activate the AbstractPPLFiniteDifferencesExt implementation.",
+        ),
+    )
+end
+
+function _assert_gradient_capability(prepared)
+    prepared isa AbstractPrepared{:jacobian} && throw(
```
penelopeysm:
```suggestion
_assert_gradient_capability(x) = nothing
function _assert_gradient_capability(::AbstractPrepared{:jacobian})
    throw(...)
end
```

ext/AbstractPPLLogDensityProblemsExt.jl
```diff
@@ -0,0 +1,43 @@
...
+function LogDensityProblems.capabilities(::VectorEvaluator{C,true}) where {C}
+    return LogDensityProblems.LogDensityOrder{1}()
+end
+
+function LogDensityProblems.logdensity_and_gradient(p::AbstractPrepared{:gradient}, x)
+    return AbstractPPL.value_and_gradient(p, x)
+end
+function LogDensityProblems.logdensity_and_gradient(e::VectorEvaluator{C,true}, x) where {C}
+    return AbstractPPL.value_and_gradient(e, x)
+end
```
penelopeysm: I feel like you shouldn't need to worry about whether it's possible to calculate the gradients in these methods. It would be simpler to just forward it to the `value_and_gradient` call and if it can't calculate the gradient, then let `value_and_gradient` error

ext/AbstractPPLLogDensityProblemsExt.jl
```diff
@@ -0,0 +1,43 @@
...
+
+function LogDensityProblems.dimension(p::AbstractPrepared)
+    return LogDensityProblems.dimension(p.evaluator)
+end
+LogDensityProblems.dimension(e::VectorEvaluator) = AbstractPPL.ADProblems.dimension(e)
+function LogDensityProblems.dimension(e::NamedTupleEvaluator)
+    return AbstractPPL.Utils.flat_length(e.inputspec)
+end
+
+function LogDensityProblems.capabilities(p::AbstractPrepared{:gradient})
```
penelopeysm: I believe `capabilities` is supposed to be defined on the types not the values. The default definition for `capabilities` on a value is to look up the type:

https://github.com/tpapp/LogDensityProblems.jl/blob/b4f1f4dd2190b16153f72fa2ff9e96dd34ba574c/src/LogDensityProblems.jl#L81-L83

The current implementation means that if you call `capabilities` on a value it will work fine, but it will fail on a type.

See also https://github.com/TuringLang/DynamicPPL.jl/blob/4201c6ce13c9d5a7d1756ab2a4f9c2564781587e/src/logdensityfunction.jl#L572-L579

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+function (e::NamedTupleEvaluator{true})(values::NamedTuple)
+    _assert_namedtuple_shape(e, values)
+    return e.f(values)
+end
+(e::NamedTupleEvaluator{false})(values::NamedTuple) = e.f(values)
+
+# Two separate overloads rather than one `(::VectorEvaluator)(::AbstractVector{<:Integer})`
+# to avoid ambiguity with `(::VectorEvaluator{true})(::AbstractVector)`.
+(e::VectorEvaluator{true})(x::AbstractVector{<:Integer}) = throw(MethodError(e, (x,)))
+(e::VectorEvaluator{false})(x::AbstractVector{<:Integer}) = throw(MethodError(e, (x,)))
```
penelopeysm: Is there a reason why these methods are needed? They define a method only to throw a MethodError (which is what happens if there is no method)?

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+        ArgumentError(
+            "`mode=:jacobian` requires `f(x)` to return an AbstractVector; got $(typeof(val)).",
+        ),
+    )
+    return (val, similar(x, length(val), 0))
+end
+
+"""
+    NamedTupleEvaluator{Checked}(f, prototype)
+    NamedTupleEvaluator(f, prototype)  # equivalent to `NamedTupleEvaluator{true}(f, prototype)`
```
penelopeysm:
```suggestion
    NamedTupleEvaluator{Checked}(f, prototype::NamedTuple)
    NamedTupleEvaluator(f, prototype::NamedTuple)
```

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+Internal evaluator shape for scalar functions of a `NamedTuple` input with a
+stable prototype. Used by AbstractPPL's AD extensions; this is not part of the
+public API.
+
+`Checked` controls whether the call method validates that an input `NamedTuple`
+has the same type as the prototype captured during preparation.
+"""
+struct NamedTupleEvaluator{Checked,F,P<:NamedTuple}
+    f::F
+    inputspec::P
```
penelopeysm, line 98:
```suggestion
    prototype::P
```
for consistency?

.github/workflows/CI.yml
```diff
@@ -6,49 +6,69 @@ on:
...
+          - '1'
+          - 'lts'
+    steps:
+      - uses: actions/checkout@v6
+      - uses: julia-actions/setup-julia@v2
+        with:
+          version: ${{ matrix.version }}
+      - uses: julia-actions/cache@v3
+      - uses: julia-actions/julia-buildpkg@v1
+      - run: julia --eval 'include("test/run_ext_tests.jl")'
```
penelopeysm: `julia --project=. test/run_ext_tests.jl` would do the same thing a bit cleaner

.github/workflows/CI.yml
```diff
@@ -6,49 +6,69 @@ on:
...
+        label:
+          - forward_diff
+          - mooncake
+          - enzyme
+          - differentiation_interface
+          - finite_differences
+          - logdensityproblems
+        version:
+          - '1'
+          - 'lts'
```
penelopeysm: is there a reason to use `lts` instead of `min` which the original CI job uses?

.github/workflows/CI.yml
```diff
@@ -6,49 +6,69 @@ on:
...
       - uses: codecov/codecov-action@v5
         with:
           files: lcov.info
           token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: true
+
+  ext:
+    name: Ext (${{ matrix.label }}, ${{ matrix.version }})
+    runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.label == 'enzyme' }}
```
penelopeysm: you already have fail-fast: false so this shouldn't be needed

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
+module ADProblems
+
+using ADTypes: ADTypes
```
penelopeysm: Is this dep actually used in `src/`? I couldn't see any usage. In which case, this doesn't seem like it's needed here or in `Project.toml`.

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+# callable from the problem. Downstream packages (e.g. DynamicPPL) pass
+# already-callable objects, so the safe default is to return them unchanged.
+prepare(problem, values::NamedTuple) = problem
+prepare(problem, x::AbstractVector{<:AbstractFloat}) = problem
+
+# Each backend owns `check_dims` and `mode` on its own signature: kwarg dispatch
+# follows the most specific positional method, so this fallback cannot forward to
+# a backend method that doesn't itself accept these kwargs.
+function prepare(
+    adtype,
```
penelopeysm: Like if this was `adtype::AbstractADType` then I could see us needing the ADTypes dep. But right now that's not the case

src/ADProblems.jl
```diff
@@ -0,0 +1,295 @@
...
+    x::AbstractVector{<:AbstractFloat};
+    check_dims::Bool=true,
+    mode::Symbol=:gradient,
+)
+    _check_mode(mode)
+    throw(
+        ArgumentError(
+            "`prepare($(nameof(typeof(adtype)))(), ...)` requires loading the corresponding AD backend.",
+        ),
+    )
```
penelopeysm: These methods can be handled with error hints too, see e.g. https://github.com/JuliaDiff/DifferentiationInterface.jl/blob/5a3ea3845c90aa69d4d23720f5c299af25268a9a/DifferentiationInterface/src/init.jl#L1-L28 for some ideas perhaps?

penelopeysm (commented):
As requested, I've not looked at the AD extensions. The biggest comments I have are

- Consider moving to another package please
- Docs would be really useful.

yebai:
Thanks for the review. Changes made:
                                                                                                                            
  **`AbstractFloat` → `Real`** — applied across `src/ADProblems.jl` and all AD extension files.                             
                                                                            
  **Removed explicit derivative mode** — `prepare` no longer takes `mode=:gradient/:jacobian`. The derivative kind is       
  inferred from the prepared function’s prototype output:                   
  - scalar output → use `value_and_gradient`                                                                                
  - vector output → use `value_and_jacobian`                                                                                
  - calling the wrong derivative API gives a direct `ArgumentError`.                                                        
                                                                                                                            
  **Dropped `AbstractPPL.ADProblems.dimension`** — the LDP-facing case is handled by `LogDensityProblems.dimension`.        
                                                                                                                            
  **Removed `AbstractPrepared{:gradient/:jacobian}`** — prepared evaluators now subtype plain `AbstractPrepared`. The one remaining capability query uses an internal `_supports_gradient` trait instead.
                                                                                                                            
  **LDP capabilities** — avoided evaluating user code inside `LogDensityProblems.capabilities(::AbstractPrepared)`.         
  Type-level capabilities still default conservatively to `LogDensityOrder{0}()` per the LDP convention; value-level methods expose gradient capability where the prepared object can actually support it.                                                
                                                                                                                            
  **Error handling** — removed throwing fallback methods for `prepare(adtype, ...)`, `value_and_gradient`, and              
  `value_and_jacobian` where they could cause method-ambiguity issues; missing backend guidance is handled with
  `register_error_hint` in `__init__`. `test_autograd`’s missing-backend throw changed from `ArgumentError` to `error()`.   
  The integer-vector `throw(MethodError(...))` overloads on `VectorEvaluator` are intentionally kept because otherwise
  `AbstractVector{Int}` silently matches the `AbstractVector` call method.

  **`Checked` → `Validate`** — renamed throughout both evaluator types and their docstrings, including a note on the `Trivial` type-stability implication.
                                                                                                                            
  **NT inputs in LDP ext** — removed `NamedTupleEvaluator`-specific `logdensity`, `dimension`, and `capabilities` overloads; NT-backed evaluators should be used via `value_and_gradient(prepared, values::NamedTuple)` directly.
                                                                                                                            
  **ADTypes dep** — removed `using ADTypes: ADTypes` from `src/`; kept in `[deps]` since the extension files import from it.
   Aqua’s stale-deps check is configured to ignore it.
                                                                                                                            
  **Docs** — added a dedicated evaluator/AD docs page covering structural vs AD preparation, vector inputs, NamedTuple      
  inputs, Jacobians, non-AD structural preparation, `test_autograd`, and supported backends.
                                                                                                                            
  **Docstring typo** (line 28), **CI** (`lts` → `min`, cleaner test command) — fixed.                                       
   
  ---                                                                                                                       
                                                                            
  Not changed:                                                                                                              

  - **LDP as weak dep** — kept as an extension because it is only needed for interop and does not need to be loaded for the core API.                                                                                                                                                                                
  - **`test_autograd` generalisation** — deferred; FiniteDifferences-only for now.
  - **CI `continue-on-error`** — not redundant with `fail-fast: false`: the former prevents Enzyme failures from failing the overall workflow check; the latter prevents other matrix jobs from being cancelled. Both are needed.

penelopeysm:
Could you (or Claude) explain why you are rejecting the parts of the review that you rejected?

penelopeysm:
(I think it's important for future readers to understand why the decisions were made that way)

yebai:
Thanks, Penny. Docs have been added. All other raised issues should be addressed or explained in the comments above.      
                                                                                                                            
  As this will likely only be implemented within TuringLang, I don't think registering a new package for it would be worth  
  it. I created AbstractPPL specifically for functionalities like those in this PR.

EDIT: fixed a few edge cases for AbstractPPLDIExt (1) AutoReverseDiff needs to respect the provided compiled tag (2) AutoEnzyme need `DI.Const` for evaluator objects.

.github/workflows/CI.yml
```diff
@@ -6,49 +6,66 @@ on:
...
```
penelopeysm: I still think that continue-on-error is not the right thing to do here, and I don't think that the explanation from Claude is relevant either.

> CI continue-on-error — not redundant with fail-fast: false: the former prevents Enzyme failures from failing the overall workflow check; the latter prevents other matrix jobs from being cancelled. Both are needed.

There isn't really a 'overall workflow check' for a GitHub workflow (each individual matrix combination is its own entry) and even if there was, the right thing to do is to inspect the failures when things fail.

CI runs a series of tests, each of which can individually fail ❌ or succeed ✅. If you set `continue-on-error` here what it means is that, _regardless of whether the test actually succeeds or not_, it will always display as a green tick. What that means in practice is that you, or the unfortunate soul who has to maintain this, will never look into the CI logs to see whether it passes or not. In other words it just gives you false confidence in whether the tests pass. It would arguably be more honest to just not even run the test.

If there are known failures, you could easily overrule it by just ignoring it, which happens quite often on real projects (for example some of Turing's CI jobs segfault due to Julia bugs and there's nothing I can do about it, so I just ignore it).

Alternatively, and even better, use `@test_broken` to mark things that are known to fail. That way when they start to pass you will know.

Project.toml
```diff
@@ -6,6 +6,7 @@ desc = "Common interfaces for probabilistic programming"
 version = "0.14.2"
 
 [deps]
+ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
```
penelopeysm, line 9: Claude is also wrong about keeping this as a dep. If you need something that is only loaded in the extensions, it should go under weakdeps.

Project.toml
```diff
@@ -18,18 +19,25 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
...
 
 [weakdeps]
+DifferentiationInterface = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+LogDensityProblems = "6fdf6af0-433a-55f7-b3ed-c6c6e0b8df7c"
 
 [extensions]
+AbstractPPLDifferentiationInterfaceExt = ["DifferentiationInterface"]
 AbstractPPLDistributionsExt = ["Distributions", "LinearAlgebra"]
+AbstractPPLLogDensityProblemsExt = ["LogDensityProblems"]
```
penelopeysm, line 29: And then you can add it to the list of extension triggers which will allow you to `using ADTypes` in the extension.

```suggestion
AbstractPPLDifferentiationInterfaceExt = ["DifferentiationInterface", "ADTypes"]
```

However, note that DI reexports the entirety of ADTypes, so do you really even need the import?

test/Aqua.jl
```diff
@@ -3,6 +3,6 @@ module AquaTests
 using Aqua: Aqua
 using AbstractPPL
 
-Aqua.test_all(AbstractPPL)
+Aqua.test_all(AbstractPPL; stale_deps=(; ignore=[:ADTypes]))
```
penelopeysm: Claude is also wrong to disable this test! If you don't make ADTypes a dep, it won't be stale, and won't need to be ignored.

penelopeysm (changes requested):

src/utils.jl
```diff
@@ -0,0 +1,139 @@
...
+# Always allocates: `_unflatten` calls `similar` for each array field. Gains from
+# buffer reuse are negligible relative to gradient computation cost.
+function unflatten_to!!(x, buf::AbstractVector)
+    n = flat_length(x)
+    length(buf) == n || throw(
+        DimensionMismatch("Expected a vector of length $n, but got length $(length(buf))."),
+    )
+    value, _ = _unflatten(x, buf, 1)
+    return value
+end
```
penelopeysm, line 145: This function is not type-stable for NamedTuples which will probably translate into perf issues:

```julia
julia> using AbstractPPL

julia> @inferred AbstractPPL.Utils.unflatten_to!!((a=1, b=2, c=3), zeros(3))
ERROR: return type @NamedTuple{a::Float64, b::Float64, c::Float64} does not match inferred return type NamedTuple

julia> @inferred AbstractPPL.Utils.unflatten_to!!((a=1, b=[2, 3], c=3), zeros(4))
ERROR: return type @NamedTuple{a::Float64, b::Vector{Float64}, c::Float64} does not match inferred return type NamedTuple
```

It's very likely that you need a generated function to ensure type stability for NamedTuples.

Also tests to check type stability for the various inputs would probably be a good thing. There are loads of examples in DynamicPPL's VNT tests which might be helpful inspiration

.github/workflows/CI.yml
yebai: Good to know that -- thanks @penelopeysm

src/utils.jl
yebai: Good catch, here, too.

Project.toml
yebai: `ADTypes` is actually needed as a hard dependency for robust error handling -- it is on the fence, but I think it is useful as a hard dependency as it simplifies weak deps triggering, eg, ForwardDiff extension only needs `using AbstractPPL, ForwardDiff`.

yebai:
Superseded by https://github.com/TuringLang/AbstractPPL.jl/pull/157

PR #157, opened 2026-04-30
Add Evaluators: prepare interface, and vectorisation utilities

yebai:
Introduces `AbstractPPL.Evaluators`, a new submodule that provides the structural plumbing for preparing and calling model evaluators, with or without AD.

  ## Core functions                                                                                                                                                                                 
  - `prepare(problem, x; check_dims=true)` / `prepare(adtype, problem, x; check_dims=true)` — structural and AD-aware preparation; AD-backend extensions implement the three-argument form                    
  - `value_and_gradient!!(prepared, x)` / `value_and_jacobian!!(prepared, x)` — stub interface for derivative computation; returned arrays may alias internal cache buffers
  - `evaluate!!(evaluator, x)` — extends the existing `AbstractPPL.evaluate!!` for `Prepared`, `VectorEvaluator`, and `NamedTupleEvaluator`                                                                   
                                                                                                                                                                                                              
  ## Evaluators                                                                                                                                                                                         
  - `VectorEvaluator{CheckInput}` — wraps a callable with a flat-vector input contract; `CheckInput=false` skips the per-call length check for AD hot paths                                                   
  - `NamedTupleEvaluator{CheckInput}` — same for `NamedTuple` inputs with a fixed prototype                                                                                                                   
                                                                                                                                                                                                              
  ## Utilities                                                                                                                                                                                                
  - `flatten_to!!(buf, x)` / `unflatten_to!!(x, buf)` — round-trip vectorisation for `Real`, `Complex`, `AbstractArray`, `Tuple`, and `NamedTuple`; `@generated` NamedTuple unflatten preserves type          
  inferability                                                                                                                                                                                                
   
  ## LogDensityProblems extension                                                                                                                                                                  
  - `logdensity`, `dimension`, `capabilities` for `Prepared` and `VectorEvaluator`
  - `logdensity_and_gradient` copies the gradient out of the aliasable `!!` buffer


yebai:
@penelopeysm, this PR keeps only the vector-/NamedTuple-typed evaluator in #155 and removes all autograd-related extensions (which can be added later).  This should be ready for a final look.

yebai:
@shravanngoswamii can you also help review this?

penelopeysm:
I'll take a more thorough look later. For my own edification, what's the relationship between this and #155? Is this meant to be "Part 1" of that PR (in which case that could be closed?) or something else?

yebai:
This PR should supersede #155. Later PRs will add autograd extensions.

test/run_extras.jl
```diff
@@ -0,0 +1,15 @@
...
+const TEST_SUBDIRS = (logdensityproblems="ext",)
+const VALID_LABELS = string.(keys(TEST_SUBDIRS))
+
+label = get(ENV, "LABEL", nothing)
+label === nothing && error("Set LABEL to one of: $(join(VALID_LABELS, ", "))")
+label in VALID_LABELS ||
+    error("Unknown LABEL=$label. Valid options: $(join(VALID_LABELS, ", "))")
+
+subdir = TEST_SUBDIRS[Symbol(label)]
+include(joinpath(@__DIR__, subdir, label, label * ".jl"))
```
penelopeysm: Why not

- use the string "ext/logdensityproblems" as the label (this mapping is rather overengineered imo)
- fix the name of the test file to be `ext/logdensityproblems/main.jl`, then you could do `include(joinpath(@__DIR__, label, "main.jl"))`?

src/evaluators/utils.jl
```diff
@@ -0,0 +1,143 @@
...
+    if isempty(Names)
+        return :((NamedTuple(), offset))
+    end
+    block = Expr(:block, :(off = offset))
+    val_syms = Symbol[]
+    for name in Names
+        v = gensym(name)
+        push!(val_syms, v)
+        push!(
+            block.args, :(($v, off) = _unflatten(getfield(x, $(QuoteNode(name))), buf, off))
```
penelopeysm: Instead of `getfield(...)` you can use `x[$name]` which is easier to read.

src/evaluators/utils.jl
```diff
@@ -0,0 +1,143 @@
+# Vectorisation utilities
+
+# This utility only supports a small structural subset so flattening stays
+# predictable and reconstruction can use `x` as the template.
+
+flat_length(x::Union{Real,Complex}) = 1
+flat_length(x::AbstractArray{<:Union{Real,Complex}}) = length(x)
+flat_length(x::Tuple) = mapreduce(flat_length, +, x; init=0)
+flat_length(x::NamedTuple) = mapreduce(flat_length, +, values(x); init=0)
```
penelopeysm: this could be simplified to `sum(flat_length, x)`

docs/src/evaluators.md
```diff
@@ -0,0 +1,138 @@
...
+`Prepared` has three fields: `adtype`, `evaluator` (the user-facing callable),
+and `cache` (backend-specific pre-allocated state such as ForwardDiff configs or
+Mooncake tapes). Backend extensions dispatch on the cache type:
+
+```julia
+function AbstractPPL.prepare(
+    adtype::MyADType, problem, x::AbstractVector{<:Real}; check_dims::Bool=true
+)
+    f = # extract callable from problem
+        cache = MyCache(f, x)
```
penelopeysm: this code is weird, didn't know ai can make typos?

docs/src/evaluators.md
```diff
@@ -0,0 +1,138 @@
...
+
+AbstractPPL provides a small interface for preparing callables and asking a
+prepared evaluator for values and derivatives. `prepare` binds a callable to a
+sample input that establishes the expected input shape and type;
+`value_and_gradient!!` and `value_and_jacobian!!` then return the value and
+derivative together.
+
+The `!!` suffix signals that the returned gradient or Jacobian **may alias
+internal cache buffers** of the prepared evaluator. Copy if you need to retain
+the result past the next call.
```
penelopeysm: is it possible to have an example demonstrating what happens if the user is not careful with this?

docs/src/evaluators.md
```diff
@@ -0,0 +1,138 @@
...
+    p::Prepared{<:AbstractADType,<:VectorEvaluator,<:MyCache}, x::AbstractVector{<:Real}
+)
+    # use p.cache to avoid allocations
+    return ...
+end
+```
+
+Pass `check_dims=false` in your `prepare` implementation to construct a
+`VectorEvaluator{false}`, which skips the per-call length check. AD libraries
+that guarantee input shape (ForwardDiff, Mooncake, etc.) should do this to
```
penelopeysm: maybe i am missing something but how does the AD backend guarantee input shape -- isn't it the responsibility of the caller to pass input of the right shape?

docs/src/evaluators.md
```diff
@@ -0,0 +1,138 @@
...
+
+The two-argument form `prepare(problem, x)` is available without any AD package.
+It returns the callable unchanged by default, so code that calls `prepare`
+unconditionally works regardless of which backends are loaded:
+
+```@example ad
+sumsimple(x) = sum(x)
+p = prepare(sumsimple, zeros(3))
+p([1.0, 2.0, 3.0])
+```
```
penelopeysm, line 155: is there a purpose for this no-AD interface considering that one could just call `sumsimple` directly to skip the faff?

src/evaluators/Evaluators.jl
```diff
@@ -0,0 +1,217 @@
...
+
+function (e::NamedTupleEvaluator{true})(values::NamedTuple)
+    _assert_namedtuple_shape(e, values)
+    return e.f(values)
+end
+(e::NamedTupleEvaluator{false})(values::NamedTuple) = e.f(values)
+
+# Reject integer vectors with a clear error rather than letting them flow into
+# AD backends (which usually fail confusingly). Split per `CheckInput` to avoid
+# an ambiguity with the `(::VectorEvaluator{true})(::AbstractVector)` method above.
```
penelopeysm: to avoid this you can just insert a check inside the other methods

```julia
if x isa AbstractVector{<:Integer}
    error(...)
end
```

or alternatively make the method take an argument of `::AbstractVector{T}` and then do `if T <: Integer`.

the compiler will be able to optimise it away for inputs that are not integers

src/evaluators/Evaluators.jl
```diff
@@ -0,0 +1,217 @@
...
+function _reject_integer_input(::VectorEvaluator, x)
```
penelopeysm: this function doesn't need the first argument

test/evaluators/utils.jl
```diff
@@ -0,0 +1,74 @@
...
+        vz = flatten_to!!(nothing, z)
+        @test vz == z
+        @test unflatten_to!!(z, vz) == z
+    end
+
+    @testset "tuple round-trip" begin
+        x = (1.0, [2.0, 3.0], (4.0 + 1.0im,))
+        v = flatten_to!!(nothing, x)
+        @test v == ComplexF64[1.0, 2.0, 3.0, 4.0 + 1.0im]
+        @test unflatten_to!!(x, v) == x
```
penelopeysm, line 36: This test and the NamedTuple one below currently pass, but I'm not sure if they _should_:

```julia
julia> using AbstractPPL.Evaluators: flatten_to!!, unflatten_to!!

julia> x = (1.0, [2.0, 3.0], (4.0 + 1.0im,))
(1.0, [2.0, 3.0], (4.0 + 1.0im,))

julia> v = flatten_to!!(nothing, x)
4-element Vector{ComplexF64}:
 1.0 + 0.0im
 2.0 + 0.0im
 3.0 + 0.0im
 4.0 + 1.0im

julia> x2 = unflatten_to!!(x, v)
(1.0 + 0.0im, ComplexF64[2.0 + 0.0im, 3.0 + 0.0im], (4.0 + 1.0im,))
```

`x2` is not really the same thing as `x`, even if `Base.:(==)` returns `true`. For example, `@test typeof(x) == typeof(x2)` would fail.

src/evaluators/utils.jl
```diff
@@ -0,0 +1,143 @@
...
+    return buf
+end
+
+function _flatten_to!(buf::AbstractVector, x::Union{Real,Complex}, offset::Int)
+    buf[offset] = x
+    return offset + 1
+end
+
+function _flatten_to!(
+    buf::AbstractVector, x::AbstractArray{<:Union{Real,Complex}}, offset::Int
```
penelopeysm: there are some cases in base Julia which probably also need to be handled carefully here:

```julia
julia> using LinearAlgebra

julia> m = randn(2, 2); n = Symmetric(m * m')
2×2 Symmetric{Float64, Matrix{Float64}}:
 4.6407   1.10794
 1.10794  4.1

julia> fn = flatten_to!!(nothing, n)
4-element Vector{Float64}:
 4.640701063537344
 1.1079436748534797
 1.1079436748534797
 4.100003866778875

julia> n2 = unflatten_to!!(n, fn)
ERROR: ArgumentError: Cannot set a non-diagonal index in a symmetric matrix
[...]
```

also for PPL purposes I think cholesky factors also need to be handled:

```julia
julia> c = cholesky(n)
Cholesky{Float64, Matrix{Float64}}
U factor:
2×2 UpperTriangular{Float64, Matrix{Float64}}:
 2.15423  0.514311
  ⋅       1.95844

julia> flatten_to!!(nothing, c)
ERROR: ArgumentError: This value cannot be flattened into a vector.
```

src/evaluators/Evaluators.jl
```diff
@@ -0,0 +1,217 @@
...
+end
+
+function _assert_supported_output(y)
+    (y isa Number || y isa AbstractVector) || throw(
+        ArgumentError(
+            "A prepared AD evaluator must return a scalar or AbstractVector; got $(typeof(y)).",
+        ),
+    )
+    return nothing
+end
```
shravanngoswamii: These are not called anywhere in this PR

test/evaluators/utils.jl
shravanngoswamii: @penelopeysm Just out of curiosity, how do you figure out these things? Is it just the experience and abundant knowledge about Julia or practice or something else I am unaware of?

penelopeysm: It is mostly experience working on these things and having been caught out by things like this 😅.

In this case there are the basic facts that NamedTuples can have heterogeneously typed values, but the intermediate vector is concretely typed, which would make one inclined to check whether there is a loss of type information when flattening.

This sort of thing comes up in different domains, for example MCMCChains's aggressive concretisation is actually the same problem but in disguise (model variables can be heterogeneously typed, shoving them into an Array causes loss of type information).

shravanngoswamii: Ohh ohkay, thanks for the details, I will look more carefully in these ways of reviewing and directions.

yebai:
Thanks for the careful review! All comments have been addressed. 

  **`test/run_extras.jl`** (`6c9e4f2`)
  - Label simplified to `ext/logdensityproblems`; the lookup table is gone. Renamed the included file to `main.jl` so the include is now `include(joinpath(@__DIR__, label, "main.jl"))` as suggested. CI
  updated.

  **`src/evaluators/utils.jl`**
  - `flat_length` tuple/namedtuple: now `sum(flat_length, ...; init=0)` (`6c9e4f2`).
  - `_unflatten` NamedTuple generator: `getfield(x, ...)` → `x[...]` (`6c9e4f2`).
  - Heterogeneous round-trip type loss (`typeof(x2) != typeof(x)`): `unflatten_to!!` now reconstructs leaves using `x`'s types via `convert(typeof(x), buf[i])` and `similar(x)`. Round-trip preserves
  `typeof(x)`. Added an opt-in `check_eltype::Bool=false` keyword that warns when `eltype(buf) !== flat_eltype(x)`. Regression tests added (`d0ebb6c`).
  - Structured arrays (Symmetric/Hermitian/Diagonal/Tri/Bi/SymTridiagonal/AbstractTriangular): rejected up front via a `_StructuredArray` union with a clear `ArgumentError`, rather than emitting broken
  results. Cholesky/LU/QR were already rejected by the catch-all (not `<:AbstractArray`). Added a `TODO` noting that proper support for structured arrays and Cholesky factors (relevant to PPL covariance
  parameters) is a follow-up (`6c9e4f2`, `88f1661`).

  **`src/evaluators/Evaluators.jl`**
  - Integer rejection: folded into the `VectorEvaluator{true/false}` method bodies, parametrising on `T` and branching on `T <: Integer` so the compiler optimises it away for non-integer inputs. The helper
  now takes only `x` (no unused first arg) (`6c9e4f2`).
  - The unused `_assert_supported_output` / `_assert_jacobian_output` helpers: kept and given an explanatory comment — they're for AD-backend extensions to share so each `value_and_gradient!!` /
  `value_and_jacobian!!` produces a uniform error rather than rolling its own (`3ad1396`).

  **`docs/src/evaluators.md`**
  - Indentation typo in the `prepare` example fixed (`6c9e4f2`).
  - `!!` aliasing: expanded with a concrete example showing the `copy(grad)` pattern and the failure mode it prevents (`6c9e4f2`).
  - `check_dims=false` rationale rewritten — no longer attributes the input-shape guarantee to AD libraries; framed as an opt-in trust mode where the caller takes responsibility, with the AD inner-loop
  scenario as a typical use (`73c97e9`, `c2b84b2`).
  - No-AD `prepare(problem, x)` rationale: now explains the use case — callers can write `prepare(...)` uniformly without knowing whether a backend is loaded, and downstream primal-only consumers (e.g.
  log-density evaluation without gradients) accept the result either way (`c2b84b2`).

src/evaluators/utils.jl
```diff
@@ -0,0 +1,190 @@
...
+            "Structured array `$(typeof(x))` is not supported by the flatten/unflatten utilities; convert to a plain `Array` first.",
+        ),
+    )
+end
+
+flat_length(x::Union{Real,Complex}) = 1
+flat_length(x::_StructuredArray) = _reject_structured(x)
+flat_length(x::AbstractArray{<:Union{Real,Complex}}) = length(x)
+flat_length(x::Tuple) = sum(flat_length, x; init=0)
+flat_length(x::NamedTuple) = sum(flat_length, values(x); init=0)
```
penelopeysm: apologies for not being clear in the last comment but i don't think `init=0` is needed either

src/evaluators/utils.jl
```diff
@@ -0,0 +1,190 @@
...
+function _flatten_to!(buf::AbstractVector, x::Union{Real,Complex}, offset::Int)
+    buf[offset] = x
+    return offset + 1
+end
+
+function _flatten_to!(
+    buf::AbstractVector, x::AbstractArray{<:Union{Real,Complex}}, offset::Int
+)
+    n = length(x)
+    copyto!(buf, offset, x, 1, n)
```
penelopeysm, line 68: The use of `1` here will error on things that aren't one-indexed so maybe either change that parameter or use `Base.require_one_based_indexing` to catch such cases?

```julia
julia> using OffsetArrays; using AbstractPPL.Evaluators: flatten_to!!

julia> oa = OffsetArray(randn(5), 6:10)
5-element OffsetArray(::Vector{Float64}, 6:10) with eltype Float64 with indices 6:10:
  0.3247729806524606
  1.5053689538589483
  1.0372531493497965
  1.1209396408261778
 -0.010548573846252697

julia> flatten_to!!(nothing, oa)
ERROR: BoundsError: attempt to access 5-element OffsetArray(::Vector{Float64}, 6:10) with eltype Float64 with indices 6:10 at index [1:5]
```

penelopeysm (commented):
just a couple more things

src/evaluators/utils.jl
penelopeysm: possible ways of working around are to use the other signatures of `copyto!`, or use `Base.firstindex` instead of 1

yebai:
@penelopeysm / @shravanngoswamii should be ready to merge if you are happy.

shravanngoswamii:
> @penelopeysm / @shravanngoswamii should be ready to merge if you are happy.

If it's fine, can you please wait for the end of the day! I am happy to approve it afterwards, let me look more carefully at it.

penelopeysm:
I can't, in good conscience, approve this PR, because I still don't agree with the direction. If you want to merge it, you have the permissions on GitHub to do so.

yebai:
> I can't, in good conscience, approve this PR, because I still don't agree with the direction. If you want to merge it, you have the permissions on GitHub to do so.

@penelopeysm, your input on Julia software engineering is appreciated.

src/evaluators/utils.jl
```diff
@@ -0,0 +1,196 @@
...
+    Base.require_one_based_indexing(buf)
+    n = flat_length(x)
+    length(buf) == n || throw(
+        DimensionMismatch("Expected a vector of length $n, but got length $(length(buf))."),
+    )
+    if check_eltype
+        expected = flat_eltype(x)
+        eltype(buf) === expected || @warn(
+            "Buffer eltype `$(eltype(buf))` differs from `flat_eltype(x) = $expected`; reconstructing using the leaf types from `x`."
+        )
```
shravanngoswamii, line 159: When `check_eltype=true` and the buffer is wider than the template, the warning fires. But if any imaginary parts are non-zero, the crash that follows comes from deep inside `copyto!` with no connection to the warning the user sees:

```julia
julia> using AbstractPPL.Evaluators: unflatten_to!!

julia> x = [1.0, 2.0]
2-element Vector{Float64}:
 1.0
 2.0

julia> buf = ComplexF64[1.0 + 0.5im, 2.0 + 0.3im]
2-element Vector{ComplexF64}:
 1.0 + 0.5im
 2.0 + 0.3im

julia> unflatten_to!!(x, buf; check_eltype=true)
┌ Warning: Buffer eltype `ComplexF64` differs from `flat_eltype(x) = Float64`; reconstructing using the leaf types from `x`.
└ @ AbstractPPL.Evaluators C:\D\Work\Turing\vectorly-ai\AbstractPPL.jl\src\evaluators\utils.jl:190
ERROR: InexactError: Float64(1.0 + 0.5im)
Stacktrace:
  [1] Real
    @ .\complex.jl:44 [inlined]
  [2] convert
    @ .\number.jl:7 [inlined]
  [3] setindex!
    @ .\genericmemory.jl:243 [inlined]
  [4] unsafe_copyto!(dest::Memory{Float64}, doffs::Int64, src::Memory{ComplexF64}, soffs::Int64, n::Int64)
    @ Base .\genericmemory.jl:153
  [5] unsafe_copyto!
    @ .\genericmemory.jl:133 [inlined]
  [6] _copyto_impl!
    @ .\array.jl:308 [inlined]
  [7] copyto!
    @ .\array.jl:294 [inlined]
  [8] _unflatten
    @ C:\D\Work\Turing\vectorly-ai\AbstractPPL.jl\src\evaluators\utils.jl:125 [inlined]
  [9] unflatten_to!!(x::Vector{Float64}, buf::Vector{ComplexF64}; check_eltype::Bool)
    @ AbstractPPL.Evaluators C:\D\Work\Turing\vectorly-ai\AbstractPPL.jl\src\evaluators\utils.jl:194
 [10] top-level scope
    @ REPL[3]:1

julia> 
```

Maybe,warning message should say something like "will throw `InexactError` if any imaginary parts are non-zero", so the user understands why the crash happened. Right now it just says "differs from" and leaves the user to figure out things!

src/evaluators/utils.jl
```diff
@@ -0,0 +1,196 @@
...
+# entries`, so a naive round-trip is lossy or fails inside `copyto!`. Reject up
+# front with a clear error rather than emitting broken results. Cholesky/LU/QR
+# are not <:AbstractArray and already fall through to the catch-all.
+#
+# TODO: extend `flatten_to!!` / `unflatten_to!!` with proper support for
+# structured arrays (independent-entry packing) and factorisation types
+# (Cholesky in particular is needed for PPL covariance parameters).
+const _StructuredArray = Union{
+    AbstractTriangular,Bidiagonal,Diagonal,Hermitian,Symmetric,SymTridiagonal,Tridiagonal
+}
```
shravanngoswamii: `_StructuredArray` rejects `Symmetric`, `Diagonal`, etc. but `Adjoint` and `Transpose` aren't in the list, so they pass through silently. The round-trip succeeds value-wise but drops the wrapper type:

```julia
julia> using AbstractPPL.Evaluators: flatten_to!!, unflatten_to!!

julia> using LinearAlgebra

julia> M = [1.0 2.0; 3.0 4.0]
2×2 Matrix{Float64}:
 1.0  2.0
 3.0  4.0

julia> x = adjoint(M)
2×2 adjoint(::Matrix{Float64}) with eltype Float64:
 1.0  3.0
 2.0  4.0

julia> v = flatten_to!!(nothing, x)
4-element Vector{Float64}:
 1.0
 2.0
 3.0
 4.0

julia> x2 = unflatten_to!!(x, v)
2×2 Matrix{Float64}:
 1.0  3.0
 2.0  4.0

julia> x == x2  # values match...
true

julia> typeof(x) == typeof(x2) # ...but type is gone
false

julia> 
```

`similar(Adjoint(M))` gives a plain `Matrix`, stripping the wrapper. The comment at [line 175](https://github.com/TuringLang/AbstractPPL.jl/pull/157/changes#diff-2605c5472d926abc0f37571c834dc6d8017e9d330b983aeefa9f023597566c17R175) says `typeof(x2) == typeof(x)` holds after a round-trip but that's broken for `Adjoint` and `Transpose`.

shravanngoswamii (changes requested):

yebai:
All perspectives are appreciated. This PR introduces a self-contained evaluator interface that is reusable across DynamicPPL, Bijectors, and AdvancedVI, and cleanly separates primal evaluation from automatic differentiation interfaces, which will be introduced in subsequent PRs.

src/evaluators/Evaluators.jl
```diff
@@ -0,0 +1,275 @@
...
+
+# Default: wrap the callable in the appropriate evaluator so per-call shape
+# checks fire even without a backend-specific `prepare` method. Downstream
+# packages (e.g. DynamicPPL) override these for their problem types.
+function prepare(problem, values::NamedTuple; check_dims::Bool=true)
+    return NamedTupleEvaluator{check_dims}(problem, values)
+end
+function prepare(problem, x::AbstractVector{<:Real}; check_dims::Bool=true)
+    return VectorEvaluator{check_dims}(problem, length(x))
+end
```
shravanngoswamii, line 69: Just to note out, isn't it a breaking change? Any package that didn't override (because the old passthrough was safe) now needs to add one...

src/evaluators/Evaluators.jl
```diff
@@ -0,0 +1,275 @@
...
+    Base.Experimental.register_error_hint(MethodError) do io, exc, args, kwargs
+        # `args` are argument types, not values (see `Base.Experimental.show_error_hints`).
+        if exc.f === prepare && length(args) >= 1 && args[1] <: AbstractADType
+            print(
+                io,
+                "\nCalling `prepare` with an AD backend requires loading the corresponding extension (e.g., `using DifferentiationInterface`).",
+            )
+        end
+    end
+end
```
shravanngoswamii, line 261: Loading DI after seeing this hint does nothing:
```julia
julia> using AbstractPPL, ADTypes

julia> AbstractPPL.prepare(ADTypes.AutoEnzyme(), "mymodel", zeros(3))
ERROR: MethodError: no method matching prepare(::AutoEnzyme{Nothing, Nothing}, ::String, ::Vector{Float64})
The function `prepare` exists, but no method is defined for this combination of argument types.
Calling `prepare` with an AD backend requires loading the corresponding extension (e.g., `using DifferentiationInterface`).

Closest candidates are:
  prepare(::Any, ::AbstractVector{<:Real}; check_dims)
   @ AbstractPPL C:\D\Work\Turing\vectorly-ai\AbstractPPL.jl\src\evaluators\Evaluators.jl:67
  prepare(::Any, ::NamedTuple; check_dims)
   @ AbstractPPL C:\D\Work\Turing\vectorly-ai\AbstractPPL.jl\src\evaluators\Evaluators.jl:64

Stacktrace:
 [1] top-level scope
   @ REPL[2]:1

julia> using DifferentiationInterface

julia> AbstractPPL.prepare(ADTypes.AutoEnzyme(), "mymodel", zeros(3))
ERROR: MethodError: no method matching prepare(::AutoEnzyme{Nothing, Nothing}, ::String, ::Vector{Float64})
The function `prepare` exists, but no method is defined for this combination of argument types.
Calling `prepare` with an AD backend requires loading the corresponding extension (e.g., `using DifferentiationInterface`).

Closest candidates are:
  prepare(::Any, ::AbstractVector{<:Real}; check_dims)
   @ AbstractPPL C:\D\Work\Turing\vectorly-ai\AbstractPPL.jl\src\evaluators\Evaluators.jl:67
  prepare(::Any, ::NamedTuple; check_dims)
   @ AbstractPPL C:\D\Work\Turing\vectorly-ai\AbstractPPL.jl\src\evaluators\Evaluators.jl:64

Stacktrace:
 [1] top-level scope
   @ REPL[4]:1

julia> 
```

src/evaluators/Evaluators.jl
```diff
@@ -0,0 +1,275 @@
...
+was constructed with `CheckInput=false`.
+"""
+function _assert_namedtuple_shape(e::NamedTupleEvaluator{true}, values)
+    typeof(values) === typeof(e.inputspec) || throw(
+        ArgumentError(
+            "Expected the same NamedTuple structure that was used to prepare this evaluator.",
+        ),
+    )
+    _shapes_match(values, e.inputspec) || throw(
+        ArgumentError(
```
shravanngoswamii, line 205: The VectorEvaluator docstring explains `CheckInput=false` is needed for AD libraries but the NamedTupleEvaluator docstring does not, even though for NamedTupleEvaluator the consequence is worse: the typeof check always fails with dual types, see:
```julia
julia> using AbstractPPL.Evaluators: NamedTupleEvaluator

julia> ne = NamedTupleEvaluator(nt -> nt.a^2 + sum(nt.b), (a=0.0, b=zeros(2)))
NamedTupleEvaluator{true, var"#5#6", @NamedTuple{a::Float64, b::Vector{Float64}}}(var"#5#6"(), (a = 0.0, b = [0.0, 0.0]))

julia> ne((a=1.0f0, b=Float32[2.0, 3.0])) 
ERROR: ArgumentError: Expected the same NamedTuple structure that was used to prepare this evaluator.
Stacktrace:
 [1] _assert_namedtuple_shape(e::NamedTupleEvaluator{…}, values::@NamedTuple{…})
   @ AbstractPPL.Evaluators C:\D\Work\Turing\vectorly-ai\AbstractPPL.jl\src\evaluators\Evaluators.jl:193
 [2] (::NamedTupleEvaluator{true, var"#5#6", @NamedTuple{…}})(values::@NamedTuple{a::Float32, b::Vector{…}})
   @ AbstractPPL.Evaluators C:\D\Work\Turing\vectorly-ai\AbstractPPL.jl\src\evaluators\Evaluators.jl:178
 [3] top-level scope
   @ REPL[3]:1
Some type information was truncated. Use `show(err)` to see complete types.

julia> 
```

src/evaluators/utils.jl
```diff
@@ -0,0 +1,175 @@
+# Vectorisation utilities
+
+# Opt-in: `Array` and `SubArray` are the only `AbstractArray` subtypes that
+# round-trip cleanly through `similar` + `copyto!` while preserving structure.
```
shravanngoswamii: Just a minor one:

similar(SubArray) returns a plain Array, not a SubArray and the wrapper type is dropped.

```julia
julia> v = @view([1.0, 2.0, 3.0][2:3])  
2-element view(::Vector{Float64}, 2:3) with eltype Float64:
 2.0
 3.0

julia> typeof(v)   
SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}

julia> typeof(similar(v)) 
Vector{Float64} (alias for Array{Float64, 1})

julia> using AbstractPPL.Evaluators: flatten_to!!, unflatten_to!!

julia> rebuilt = unflatten_to!!((x=v,), flatten_to!!(nothing, (x=v,)))
(x = [2.0, 3.0],)

julia> typeof(rebuilt.x) == typeof(v)
false

julia> 
```

src/evaluators/Evaluators.jl
yebai: It's not breaking — both prepare methods are new on this branch (AbstractPPL didn't have a prepare before this PR), so there's no prior passthrough behaviour for downstream packages to rely on. The wrapping default is there for consistency: `prepare(problem, x)` always returns a callable evaluator with the same shape-check contract, regardless of whether an AD backend is loaded.

shravanngoswamii (approved):

src/evaluators/Evaluators.jl
shravanngoswamii: Okay, thanks for clarification!

PR #163, opened 2026-05-19
Add order=2 Hessian preparation via value_gradient_and_hessian!!

yebai:
Extends `prepare(adtype, problem, x; order=2)` to build Hessian machinery for scalar-valued problems on the DI and Mooncake extensions, returning `(value, gradient, hessian)` from a new `value_gradient_and_hessian!!` generic. Unifies the per-extension caches (`DICache`, `MooncakeCache`) so one struct carries every derivative order, with explicit cross-arity error messages replacing prior `MethodError`s. DI uses the in-place `DI.value_gradient_and_hessian!` with caller-owned buffers; Mooncake uses its native `prepare_hessian_cache` API.

Hessian is needed by some variational inference algorithms. See https://github.com/TuringLang/AdvancedVI.jl/pull/255

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "7a57a42e-76ec-4ea3-a279-07e840d6d9cf"
 keywords = ["probabilistic programming"]
 license = "MIT"
 desc = "Common interfaces for probabilistic programming"
-version = "0.15"
+version = "0.15.1"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/docs/src/evaluators.md
+++ b/docs/src/evaluators.md
@@ -183,10 +183,11 @@ prepared = prepare(adtype, quadratic, zeros(3); order=2)
 entry handles it.
 
 Calling `value_gradient_and_hessian!!` on an `order=1` prep throws an
-`ArgumentError` — re-prepare with `order=2` instead. Likewise, calling
-`value_and_gradient!!` or `value_and_jacobian!!` on an `order=2` prep is
-unsupported; use `value_gradient_and_hessian!!` and discard the unused
-return value.
+`ArgumentError` — re-prepare with `order=2` instead. The reverse is allowed:
+`value_and_gradient!!` on an `order=2` prep returns `(value, gradient)`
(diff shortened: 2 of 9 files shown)


shravanngoswamii (commented):
Just one small suggestion.

Can we validate `order` before building the evaluator?

Right now `order=3` still calls `problem` once, and the user can get an output-arity error before the actual invalid-order error. See:

```julia
julia> using AbstractPPL, ADTypes, DifferentiationInterface, ForwardDiff

julia> calls = Ref(0);

julia> f(x) = (calls[] += 1; [1 2; 3 4]);

julia> AbstractPPL.prepare(AutoForwardDiff(), f, zeros(2); order=3)
ERROR: ArgumentError: A prepared AD evaluator must return a scalar or AbstractVector; got Matrix{Int64}.
Stacktrace:
 [1] _ad_output_arity
   @ ~/Work/vectorly-ai/AbstractPPL.jl/src/evaluators/Evaluators.jl:277 [inlined]
 [2] prepare(adtype::AutoForwardDiff{…}, problem::Function, x::Vector{…}; check_dims::Bool, context::Tuple{}, order::Int64)
   @ AbstractPPLDifferentiationInterfaceExt ~/Work/vectorly-ai/AbstractPPL.jl/ext/AbstractPPLDifferentiationInterfaceExt.jl:75
 [3] top-level scope
   @ REPL[5]:1
Some type information was truncated. Use `show(err)` to see complete types.

julia> calls[]
1

julia>
```

I expected this to throw ``ArgumentError("`order` must be 1 or 2, got 3.")`` without calling `f`.

Maybe both DI and Mooncake paths can do:

```julia
order in (1, 2) || throw(ArgumentError("`order` must be 1 or 2, got $order."))
```

before the `AbstractPPL.prepare(problem, ...)` / `evaluator(x)`.

yebai:
Thanks for catching this! Fixed in b98f027, together with a few related improvements helpful for AdvancedHMC/AdvancedVI: 

  - **`order` is a type parameter of `Prepared`** with a public `order(::Prepared)` accessor, so it can be retrieved reliably without inspecting cache internals.

shravanngoswamii (approved):
Looks good to me! Thank you!

PR #168, opened 2026-06-01
Add the `of` type from JuliaBUGS.jl

sunxd3:
Adds the `of` type system to AbstractPPL: a small, declarative way to describe the shape, element type, and support of model parameters as ordinary Julia types.

`of(...)` and the `@of` macro construct `OfType` specifications that double as type annotations. Only `of` and `@of` are exported; `flatten`/`unflatten` and the `OfType` subtypes are marked `public`.

Highlights:

- Scalar, bounded, array, and named-tuple specs: `of(Float64, 0, 1)`, `of(Int)`, `of(Array, 3, 4)`, `@of(mu = of(Real), sigma = of(Real, 0, nothing))`, and constants via `constant=true`.
- Symbolic dimensions and bounds — including arithmetic such as `n + 1` — resolved with `of(T; n=...)`.
- `rand`, `zero`, `size`, `length`, and `flatten`/`unflatten` over the specs; `unflatten` is AD-transparent and type-stable.

The implementation is moved and then updated from JuliaBUGS.jl.


src/AbstractPPL.jl
```diff
@@ -95,4 +95,15 @@ export AbstractOptic,
...
+export of, @of
+@static if VERSION >= v"1.11.0"
+    eval(
+        Meta.parse(
+            "public OfType, OfReal, OfInt, OfArray, OfNamedTuple, OfConstantWrapper, " *
+            "flatten, unflatten",
+        ),
+    )
+end
+
```
sunxd3, line 108: new public keywords worth reviewing

sunxd3:
@yebai @shravanngoswamii this is ready if you have time to do a review pass. Otherwise I'll merge it the day after tomorrow, given that the risk is low.

PR #177, opened 2026-07-07
Call-time `context` overrides (#167); harden reused reverse-mode caches

### LINKED ISSUE #167: Allow `value_and_gradient/jacobian/hessian!!` accept `contexts`
`contexts` is currently frozen after `AbstractPPL.prepare`. It may be useful to allow `value_and_gradient/jacobian/hessian!!` to accept `contexts` in some cases.

yebai:
In reverse mode Mooncake's order=1 path passed the problem function `f` and any `context` to Mooncake as ordinary arguments, without zeroing their cotangents. When `context` carries *differentiable* data reached through an in-place linear solve — whose reverse rule reads its argument's cotangent back — a reused prepared cache let that cotangent accumulate across calls and returned an incorrect `x`-gradient after the first one ([Mooncake #1238](https://github.com/chalk-lab/Mooncake.jl/issues/1238)). The fix closes `f` and `context` into a `NoTangent`-typed `_ADTarget` so they carry no cotangent and `x` is the only active argument, which also unifies the `AutoMooncake`/`AutoMooncakeForward` paths; a reused-prep regression test (a differentiable `context` routed through a triangular solve, checked against a fresh prep and finite differences) guards it.

It also adds a `context=` keyword to `value_and_gradient!!` / `value_gradient_and_hessian!!` (closes #167) that overrides the frozen context for a single call, without re-preparing, across all three AD extensions: the default `nothing` leaves the zero-allocation hot path unchanged, and a matching-shape `Tuple` reuses the type-keyed cache. Compiled-tape ReverseDiff and Mooncake's Hessian bake context into their prepared state and throw an `ArgumentError` on such an override (empty input always accepts one).


sunxd3 (approved):
This makes a lot of sense, thanks for the fix.

sunxd3:
@yebai I pushed a commit contains:
* `context` argument for `value_and_jacobian`
* more input validation for `_evaluate_with_context`
* minor doc clarification

commit message (https://github.com/TuringLang/AbstractPPL.jl/pull/177/changes/74aa79519b0549414f52486de18e77d3ff7601d4#diff-f34a3deb90d5817c924be448c257eeb9ab80df79744c1720173ccb1c0164d146) gives a bit more details. again, feel free to drop any of the things I added, the PR contains no major issue pre-my-commit

src/evaluators/Evaluators.jl
```diff
@@ -267,20 +278,77 @@ _check_ad_input(::VectorEvaluator{false}, ::AbstractVector) = nothing
...
-@inline _evaluate_with_context(e::VectorEvaluator, x, context::Tuple) = e.f(x, context...)
+@inline function _evaluate_with_context(
+    e::VectorEvaluator, x::AbstractVector{T}, context::Tuple
+) where {T}
+    T <: Integer && _reject_integer_input(x)
+    return e.f(x, _resolve_context(e, context)...)
+end
+# A non-`Tuple` override fails with the same `ArgumentError` as the AD paths.
+@inline _evaluate_with_context(e::VectorEvaluator, x, context) =
+    _resolve_context(e, context)
```
yebai: This method is confusing as `_resolve_context` is used in a weird way to recycle its error message. Better to explicitly error here.

src/evaluators/Evaluators.jl
```diff
@@ -267,20 +278,77 @@ _check_ad_input(::VectorEvaluator{false}, ::AbstractVector) = nothing
...
+# element type in one comparison that constant-folds away when the types match,
+# and array elements are additionally compared by `size` (top-level elements
+# only, mirroring the granularity of the backends' own cache validation). The
+# `nothing` hot path is untouched.
 @inline _resolve_context(e::VectorEvaluator, ::Nothing) = e.context
-@inline _resolve_context(::VectorEvaluator, context::Tuple) = context
+@inline function _resolve_context(e::VectorEvaluator, context::Tuple)
+    typeof(context) === typeof(e.context) ||
+        _throw_context_type_mismatch(e.context, context)
+    _check_context_sizes(e.context, context)
```
yebai: The context type and size checks were intentionally omitted because they introduce constant overhead for every gradient evaluation. The design assumes that users are responsible for providing the correct context.

Looking at the implementation, the type checks appear to incur only compile-time overhead. If the benchmarks confirm this, I’m happy to keep them.

I do think the size checks should be removed. In principle, users can pass arbitrary data as context, so enforcing size constraints is difficult to do in a general and meaningful way.

yebai (commented):
Thanks @sunxd3 -- all good catches. Some comments below. Can you take this forward?

src/evaluators/Evaluators.jl
sunxd3: thanks, makes sense. agree that size/shape checks are hard to done properly

and yes, the type check is compiled away, so no runtime performance impact

